PR #35, opened 2020-03-03
Debug logging over parallel port

crawfxrd:
Implement logging over the parallel port.

- Add a command to the Mega to put it in peripheral mode and write logs from the EC to the serial port
- Implement host device logic in EC to write logs through parallel port

### DIFF
--- a/src/board/arduino/mega2560/parallel.c
+++ b/src/board/arduino/mega2560/parallel.c
@@ -10,6 +10,8 @@
 #include <arch/uart.h>
 
 // Mapping of 24-pin ribbon cable to parallel pins. See schematic
+#define IT8587
+#if defined(IT8587)
 #define PINS \
     /* Data (KSO0 - KSO7) - bi-directional */ \
     PIN(d0, 1) \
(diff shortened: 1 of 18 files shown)


src/board/arduino/mega2560/parallel.c
```diff
@@ -228,6 +249,55 @@ int parallel_transaction(struct Parallel * port, uint8_t * data, int length, boo
...
+
+    int i;
+    uint8_t byte = 0;
+    for (i = 0; i < length; i++) {
+        // Wait for host to indicate it's written data
+        while (gpio_get(port->data_n) && gpio_get(port->addr_n)) {}
+
+        *addr = !gpio_get(port->addr_n);
+
+        if (read) {
```
jackpot51: Read/write direction is controlled by the host with the write_n line. If this line is low, then the host is writing to the peripheral.

src/board/arduino/mega2560/parallel.c
```diff
@@ -228,6 +249,55 @@ int parallel_transaction(struct Parallel * port, uint8_t * data, int length, boo
...
+            data[i] = parallel_read_data(port);
+        } else {
+            byte = data[i];
+            parallel_write_data(port, byte);
+        }
+
+        // Tell host we're processing data
+        gpio_set(port->wait_n, true);
```
jackpot51: See http://studenti.fisica.unifi.it/~carla/manuali/porta_parallela/porta_parallela.html

In either cycle, leave the data lines in a floating input state until you know the type of cycle. Wait for either addr_n or data_n to go low `while (addr_n && data_n) {}`. Then read write_n to determine the direction.

In a read cycle (write_n high), set the data lines as outputs and set them to the correct values. Set wait_n high and wait until both addr_n and data_n are high `while (!addr_n || !data_n) {}`. Set data lines as floating inputs and then set wait_n low.

>  An EPP Data Read cycle is shown in Figure 5. The host de-asserts WRITE*, places the AD8-1 lines in a high impedance state, and asserts the DSTROBE* signal. The peripheral then drives the data byte on the AD8-1 lines and re-asserts WAIT* to indicate that the data byte is valid. The host will read the data lines when it sees the WAIT* unasserted, and then de-asserts DSTROBE*. The peripheral then places the AD8-1 lines in an high impedance state and asserts WAIT* to indicate it is ready for the next cycle.

In a write cycle (write_n low), set wait_n high and wait until both addr_n and data_n are high `while (!addr_n || !data_n) {}`. Read the data lines and then set wait_n low.

>  An EPP Data Write cycle is shown in Figure 4. The host first asserts WRITE*, places the data byte on the AD8-1 lines, and asserts DSTROBE*. The peripheral de-asserts WAIT* indicating that it is ready for the data byte. The host then de-asserts DSTROBE* and the peripheral latches the data lines on the rising edge of DSTROBE*. The peripheral then indicates that it is ready for the next cycle by asserting WAIT*.

I recommend using a callback in the peripheral transaction that can perform either address or data, read or write, depending on host request.

Issue #24, opened 2020-02-21
Prevent ghost keys


jackpot51:
Detect and prevent ghost keypresses

jackpot51:
![ghosting](https://user-images.githubusercontent.com/7741236/77469470-bf684f00-6dd4-11ea-87c4-4b3e3cc91bbb.png)

jackpot51:
When a keypress happens, the EC should check if the keypress connects 2 inputs and 2 rows with 3 switches. Any time this happens, the fourth switch would invalidly register at the same time.

PR #40, opened 2020-03-31
Ignore key presses that trigger ghost keys

crawfxrd:
If a key press is detected that would trigger a ghost key, ignore the key.

Based on QMK implementation ([keyboard.c](https://github.com/qmk/qmk_firmware/blob/b892a1429d753d83d179fea26a2c7b84edab840d/tmk_core/common/keyboard.c#L129)).

Resolves: #24


crawfxrd:
I was testing it and forgot to unmark it WIP.

PR #53, opened 2020-04-14
System76 common board directory

jackpot51:
Will reduce code copying between boards


jackpot51:
This isn't the whole driver abstraction I was thinking about but it is a significant reduction in copied code. Most code has been moved to src/board/system76/common which is included in each system76 board's source tree by a makefile.

PR #56, opened 2020-04-21
Fix release of key after release of layer key

### LINKED ISSUE #55: Fn layer does not take into account previously pressed keys
When pressing a Fn-enable key, such as \` (play/pause), one of two sequences can happen:
(1) Fn is pressed, \` is pressed, \` is released, Fn is release.
(2) Fn is pressed, \` is pressed, Fn is released, \` is released. This is how I normally type.

If case (1) above is used, everything behaves correctly.
If case (2) above is used instead, the following happens:
play/pase is sent to the OS.
\` is sent to the os.

This is because there is a global layer selection instead of a per-key layer selection.
If a key is pressed (play/pause) when Fn is pressed, then that key should release as its fn key even after fn is released.
Conversely, If a key is pressed without fn depressed, and then fn is depressed after, I would also expect it to release as the non-fn key.

Effectively, this means that each key should track either what was depressed or the layer that is was on when depressed so that it can be released as the same key.

Let me know if you'd prefer that I work on a patch, however it may take me about a week before I can get to it.

jackpot51:
This fixes #55

### DIFF
--- a/src/board/system76/common/kbscan.c
+++ b/src/board/system76/common/kbscan.c
@@ -214,6 +214,7 @@ void kbscan_event(void) {
     static uint8_t kbscan_layer = 0;
     uint8_t layer = kbscan_layer;
     static uint8_t kbscan_last[KM_OUT] = { 0 };
+    static uint8_t kbscan_last_layer[KM_OUT][KM_IN] = { { 0 } };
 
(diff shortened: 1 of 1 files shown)


jackpot51:
Going to have QA approval in firmware repo

Issue #58, opened 2020-04-28
lemp9: Ghost key detection blocking normal user input


ben-dyer:
Hello :wave: new system76 user here! I am experiencing the following issue on lemp9:

For a certain set of words, I misspell them essentially 100% of the time (e.g. "list" -> "lit" or ".org" -> ".or"), which is presumably due to my style of user input being incompatible with ghost key detection rules https://github.com/system76/ec/pull/40. 

I have seen #57 and it looks clear that this is an intended behaviour, but its disruptive so I am interested in an adjustment. Is it possible to configure the ghost key detection?

(:thinking: :thought_balloon: Is it as simple as cloning, modifying `src/board/system76/lemp9/kbscan.c` and flashing `make BOARD=system76/lemp9 flash_internal`? :sweat_smile: )

crawfxrd:
The current ghost key detection is as good as it gets, based on QMK's implementation. There is no configuration.

The file has moved to `src/board/system76/common/kbscan.c`. Your flash command is correct.

ben-dyer:
@crawfxrd Ok, thanks for the info!

jackpot51:
You may also want to look into debounce configuration. It is currently set to debounce after 20ms.

ben-dyer:
@jackpot51 I had a similar idea yesterday - was able to turn ghost key presses back on and used xev to check that most of the ghost presses occured in between 20-40 ms whereas normal presses were usually more than 60ms apart.

Turning debounce ms up to 40 removed some of the presses but for the words I have looked at it removes the wrong presses "list" is "liut" instead of "liust"  and build is "buisd" ("buisld" with 20 ms debounce).

Wondering whether a custom rule for what to do with a multi-key event based on a short history of keypresses might be possible. (Maybe it will turn out this is buggy, but then again maybe it would just work)

meklort:
@ben-dyer and @jackpot51 There's another possibility that may be worth looking at:
As mentioned, when you press and hold 'lis', the s key is rejected, as you mentioned. You'll also notice that if you *release* the 'l' and 'i' keys, the 's' key is *not* sent.

Looking at the code, this is due to the following:
```
            if (kbscan_has_ghost_in_row(i, new)) {
                kbscan_last[i] = new;
                continue;
            }
```
The kbscan_last[i] = new; line causes the s key to to be marked as pressed, and as such it will never be sent until it's released.
[...]

PR #60, opened 2020-05-02
keymap: Allow ghost keys to be pressed once the ghost is removed

meklort:
- Don't record a ghost key as pressed
- Debouce on ghost release to allow ghosts to settle.
- Reduce debounce time from 20ms to 15ms.

Signed-off-by: Evan Lojewski <github@meklort.com>


meklort:
This improves ghost behavior as described in #58

PR #61, opened 2020-05-07
keymap: Improve debouncing behavior.

meklort:
- Update debouncing code to properly handle rollover.
- Update ghost debouncing to properly set the debounce start time.

### DIFF
--- a/src/board/system76/common/kbscan.c
+++ b/src/board/system76/common/kbscan.c
@@ -227,11 +227,7 @@ void kbscan_event(void) {
     // If debounce complete
     if (debounce) {
         uint32_t time = time_get();
-        //TODO: time test with overflow
-        if (time < debounce_time) {
-            // Overflow, reset debounce_time
-            debounce_time = time;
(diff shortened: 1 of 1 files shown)


meklort:
@jackpot51 The previous commit had an issue due to not setting debounce_time.
During some additional testing today, I noticed that the debounce was shorted than expected in some cases, resulting in a ghost being sent. This was due to the previous debounce time being used, instead of the current time. With debounce_time set, I see these occurrence removed / significantly reduced (see below).

I've also removed the debounce wrapping code as a simple delta performs the same (and better) function. Subtracting the current time from the previous time will ensure that the value in unwrapped.

Finally, I've reviewed the code some more, and see that there are still cases where the debounce will be wrong. It's rare, but I'm pretty sure I have triggered these. Specifically:
- The debounce time is global and not per-key as such,
- Any new keypresses during an active debounce will resulting in a *smaller* debounce than expected. This can trigger in the current use case especially since debouncing starts on both a press and a release. As a result, a press/release event can cause the second portion to not be debounced properly.

Issue #62, opened 2020-05-09
Battery charging threshold


tsitavets:
It would be great to implement battery charging threshold to increase battery lifespan when working a lot from AC. For example stop charging at 80% and start charging again at 60%. It would be nice to have keyboard shortcut to enable/disable feature and to have a possibility to change thresholds somehow (because optimal thresholds levels depend on battery usage). Also we can blink by power led to tell user is feature enabled (one time blink) /disabled (two times blink) upon pressing keyboard shortcut.

crawfxrd:
Battery charging thresholds are implemented. Currently compile-time only. Need to add ACPI support and expose it in sysfs.

crawfxrd:
Charging thresholds can now be set/queried with  `SBCT`/`GBCT` ACPI methods [\[1\]](https://github.com/system76/coreboot/pull/15).

`system76_acpi` has been updated to expose data in standard sysfs entries `/sys/class/power_supply/BAT0/charge_control_{start,end}_threshold` [\[2\]](https://github.com/pop-os/system76-acpi-dkms/pull/4).

See also: pop-os/gnome-control-center#101

PR #63, opened 2020-05-19
RFC: Enable user-configurable settings including Fn Lock, Debounce, etc

meklort:
- Feature Name: user-configuration
- Start Date: 2020-05-18
- Relevant Issues:  system76/firmware-open#78 and PRs #62 and #80 

# Summary
[summary]: #summary

The change add support for users to configure the EC behavior without having to build and flash a custom firmware image. A new flash API has been added, allowing for arbitrary read/write access to flash. A new configuration API has been added, persistently storing saved settings into an unused region of flash.

# Motivation
[motivation]: #motivation

While I am able to customize and rebuild the EC firmware, not all users are able or willing to do so, especially with the risks towards bricking their device. Additionally, any time a firmware update is release, a new customized EC binary must be generated to contain the user customization.

This change will allow users to customize the EC without the above mentioned risks, while also allowing those customization to persists across firmware updates.

# Guide-level explanation
[guide-level-explanation]: #guide-level-explanation

A user may list all configurable options within the EC by using the ectool.
```
meklort@lemur ~/g/ec (config-pull)> ./ectool.sh config
    Finished release [optimized] target(s) in 0.00s
Fn Lock: When enabled, F1-F12 will activate their default function
Min: 1
Max: 0
Current Value: 1

Keyboard Debounce Time: Debounce time in milliseconds
Min: 1000
Max: 1
Current Value: 15
```
If a user wishes to customize a setting, they can also do so using the ectool.
```
meklort@lemur ~/g/ec (config-pull)> ./ectool.sh config "Keyboard Debounce Time" 20
    Finished release [optimized] target(s) in 0.01s
meklort@lemur ~/g/ec (config-pull)> ./ectool.sh config "Keyboard Debounce Time"
    Finished release [optimized] target(s) in 0.00s
Dump "Keyboard Debounce Time"
Keyboard Debounce Time: Debounce time in milliseconds
Min: 1000
Max: 1
Current Value: 20
```

# Reference-level explanation
[reference-level-explanation]: #reference-level-explanation

Two new modules are provided with this change:
## Flash API
A new flash API is added, allowing code to read/write arbitrary locations in flash. Due to the XIP nature of the EC, this code must run out of RAM. As such, it is modeled and based off of the scratch ROM code used for firmware updates.

- A new flash application with a single entry point is added.
- The flash API is configured to use *no* globals and to place all locals on the stack
   This ensures that no memory in the standard application is mangled.
- During early init, the flash API is copied to RAM using SCAR0/SCAR2
   Note: This conflicts with the scratch ROM, however this is not an issue as firmware flashing requires a full system reboot.

## Config API
A new configuration API is added, allowing existing code to register a configuration entry with the system. Each configuration entry contains a short name, description, default, min, and, max values. Additionally, an option callback can bet set, allowing a function to be called when a configuration value is updated.

Using SMFI, all configuration entries are exported to the host. Additionally, the host is able to specify a new value, within the limits, for the configuration entry. All configuration settings are stored in the EC, allowing for the host to auto detect available parameters.

When a configuration entry is registered with the system, the flash APIs are used to update the configuration value with any user-customized settings. If no user customization is found, the default value is used instead.

When a configuration entry is updated by the user, the flash API is used to store the new value persistently. In order to allow quick updates to the settings, old values in flash are invalidated before a new setting is added. This allows new settings to be stored without having to first erase flash.

## Existing Modules

Existing modules can easily be updated to support user configurations. As an example, the keyboard module has been updated to allow configurable function lock and default state in #80 . 

# Drawbacks
[drawbacks]: #drawbacks

This code not yet complete, and it does add a couple of risks:
- ~~Currently, no code handles a full flash. If the full 4KB are used with configuration entries, no more settings can be saved. In this case, the firmware should be updated to (1) erase the sector and (2) store all known settings back into flash.~~
- This code calls between two applications. The flash API is a separate image which has no knowledge of XSEG / DSEG usage. As a result, the a developer modifying the must take special not to mangle existing registers.
- Any mistakes in the flash writing code could accidentally erase or overwrite the main EC binary, bricking the device. Special care need to be taken to ensure that a device is not bricked. Note: Additional checking could be added to the flash API to limit writes to a small window.
- The configuration APIs only support a single 32bit value. In some cases, a larger value may be desired. As an example, the keyboard layout could be stored in flash using a similar mechanism. This would allow custom keymaps. At present, this use case is not supported.
- The code currently does a linear search in flash. As the ~~4KB~~ 1KB configuration space fills up, read/write access time can be impacted. This can slow down the initial boot, as well as reconfiguration time by the user.

# Rationale and alternatives
[rationale-and-alternatives]: #rationale-and-alternatives

A couple of additional possibilities were considered including:
- Using the UEFI payload to store setting in nvram. This would allow the EC firmware to never touch flash, and instead could use the existing nvram APIs in UEFI. This would require (1) UEFI to scan nvram for all EC related variables (using a new GUID) and passing these to the EC. (2) Adding a mechanism to set the UEFI variables by first reading possible entries from the EC, then writing these to NVRAM. This means the change cannot be standalone, and so it was not really considered.
- Using a more compact format on flash could have been done. The flash usage could be minimized by only storing a  single 32bit value. This compresses the used space, however means that the location in flash cannot change. As a result, firmware updates must maintain past locations and the flash must be erased to update each value. Due to these limitations, a method supporting auto-detection and arbitrary locations in flash is supported instead. Also due to the large 128KB flash vs the (current) 32KB firmware size, the larger size seems acceptable.

# Prior art
[prior-art]: #prior-art

N/A

# Unresolved questions
[unresolved-questions]: #unresolved-questions

- What is the best way for the user to interact with this? The firmware-setup UEFI app could be updated to expose these to the user. The Pop! OS settings page could be updated instead to provide a gui for these settings instead. 

# Future possibilities
[future-possibilities]: #future-possibilities

This can be further enhanced in the future (as mentioned above) to support other configuration value types (strings, blobs, etc) instead of just 32bit integers.
A number of additional settings make sense to be implemented (I have done some, but have not cleaned them up and included them here yet). These include:
- Battery charge thresholds
- Startup keyboard back light value
- (possibly, if hardware supported) Start on lid open
- Additional keyboard customization (remap pgup/pgdwn to left/right,
- USB port settings (disable left usb port when in a sleep power state)

### DIFF
--- a/src/board/system76/common/flash/flash.mk
+++ b/src/board/system76/common/flash/flash.mk
@@ -1,0 +1,56 @@
+# Set flash ROM parameters
+FLASH_OFFSET=2048
+FLASH_SIZE=1024
+CFLAGS+=-DFLASH_OFFSET=$(FLASH_OFFSET) -DFLASH_SIZE=$(FLASH_SIZE)
+
+# Copy parameters to use when compiling flash ROM
+FLASH_INCLUDE=$(INCLUDE)
+FLASH_CFLAGS=$(CFLAGS)
+
+# Include flash source.
+FLASH_DIR=$(SYSTEM76_COMMON_DIR)/flash
+# Note: main.c *must* be first to ensure that flash_start is at the correct address
+FLASH_SRC=$(FLASH_DIR)/main.c
+FLASH_INCLUDE+=$(wildcard $(FLASH_DIR)/include/flash/*.h) $(FLASH_DIR)/flash.mk
+FLASH_CFLAGS+=-I$(FLASH_DIR)/include -D__FLASH__
+
+# Add minimal source from other directories
+FLASH_SRC+=
+
+FLASH_BUILD=$(BUILD)/flash
+FLASH_OBJ=$(patsubst src/%.c,$(FLASH_BUILD)/%.rel,$(FLASH_SRC))
+FLASH_CC=\
+	sdcc \
+	-mmcs51 \
+	--model-large \
+	--opt-code-size \
+	--acall-ajmp \
+	--code-loc $(FLASH_OFFSET) \
+	--code-size $(FLASH_SIZE) \
+	--Werror
+
+# Convert from binary file to C header
+$(BUILD)/include/flash.h: $(FLASH_BUILD)/flash.rom
+	@mkdir -p $(@D)
+	xxd -s $(FLASH_OFFSET) --include < $< > $@
+
+# Convert from Intel Hex file to binary file
+$(FLASH_BUILD)/flash.rom: $(FLASH_BUILD)/flash.ihx
+	@mkdir -p $(@D)
+	makebin -p < $< > $@
+
+# Link object files into Intel Hex file
+$(FLASH_BUILD)/flash.ihx: $(FLASH_OBJ)
+	@mkdir -p $(@D)
+	$(FLASH_CC) -o $@ $^
+
+# Compile C files into object files
+$(FLASH_OBJ): $(FLASH_BUILD)/%.rel: src/%.c $(FLASH_INCLUDE)
+	@mkdir -p $(@D)
+	$(FLASH_CC) $(FLASH_CFLAGS) -o $@ -c $<
+
+# Include flash header in main firmware
+CFLAGS+=-I$(BUILD)/include
+LDFLAGS+=-Wl -g_flash_entry=$(FLASH_OFFSET)
+INCLUDE+=$(BUILD)/include/flash.h
+SRC+=$(FLASH_DIR)/wrapper.c
--- a/src/board/system76/common/flash/main.c
+++ b/src/board/system76/common/flash/main.c
@@ -1,0 +1,169 @@
+/*
+ * Copyright (C) 2020 Evan Lojewski
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#include <stdint.h>
+#include <board/flash.h>
+
+// EC indirect flash access
+volatile uint8_t __xdata __at(0x103B) ECINDAR0;
+volatile uint8_t __xdata __at(0x103C) ECINDAR1;
+volatile uint8_t __xdata __at(0x103D) ECINDAR2;
+volatile uint8_t __xdata __at(0x103E) ECINDAR3;
+volatile uint8_t __xdata __at(0x103F) ECINDDR;
+
+#define SPI_DEVICE                  (0x70)
+#define SPI_FOLLOW_MODE             (0x0F)
+#define SPI_CHIP_SELECT             (0xFD)
+#define SPI_CHIP_DESELECT           (0xFE)
+
+#define SPI_READ_STATUS_COMMAND     (0x05)
+#define SPI_READ_COMMAND            (0x0B)
+#define SPI_WRITE_COMMAND           (0x02)
+#define SPI_WRITE_ENABLE_COMMAND    (0x06)
+#define SPI_READ_STATUS_COMMAND     (0x05)
+
+#define SPI_ERASE_SECTOR_COMMAND    (0xD7)
+
+#define SPI_STATUS_WIP              (0x01)
+
+void flash_enter_follow_mode(void);
+void flash_exit_follow_mode(void);
+void flash_wait(void);
+void flash_write_enable(void);
+
+/**
+ * Main flash API entry point.
+ *
+ * NOTE: This *must* be the first function in this file to ensure that it is placed
+ *          first in the resulting binary. This is required to ensure that address
+ *          matches the address (FLASH_OFFSET) for flash_entry in wrapper.c.
+ * NOTE: __reentrant so that parameters and temperary vairables are placed on the
+ *          stack, ensuring the main application __data variables are not stomped on.
+ * NOTE: __critical to ensure interrupts are disabled. This does mean that interrupt
+ *          such as the timer will be block until flash acccess is complete
+ */
+void flash_entry(uint32_t addr, uint8_t * data, uint8_t length, uint8_t command) __reentrant __critical {
+    // Only allow access from 64KB to 128KB.
+    if (addr < 0x10000 || addr >= 0x20000 || (addr + length) > 0x20000) return;
+
+    if (command == FLASH_COMMAND_READ) {
+        while (length) {
+            // Fast read.
+            ECINDAR3 = SPI_DEVICE;
+            ECINDAR2 = addr >> 16;
+            ECINDAR1 = addr >> 8;
+            ECINDAR0 = addr;
+
+            *data = ECINDDR;
+
+            addr++;
+            data++;
+            length--;
+        }
+    } else if (command == FLASH_COMMAND_WRITE) {
+        flash_enter_follow_mode();
+
+        while (length) {
+            // Note, this is the slow way to do it, but it's simple and all bytes are written properly.
+            flash_write_enable();
+
+            // Select the device
+            ECINDAR1 = SPI_CHIP_SELECT;
+
+            // Send write command
+            ECINDDR = SPI_WRITE_COMMAND;
+            ECINDDR = addr >> 16;
+            ECINDDR = addr >> 8;
+            ECINDDR = addr;
+
+            ECINDDR = *data;
+
+            data++;
+            length--;
+            addr++;
+
+            // Deselect
+            ECINDAR1 = SPI_CHIP_DESELECT;
+            ECINDDR  = 0x00;
+
+
+            // Wait WIP to be cleared
+            flash_wait();
+        }
+
+        flash_exit_follow_mode();
+    } else if (command == FLASH_COMMAND_ERASE_1K) {
+        flash_enter_follow_mode();
+
+        flash_write_enable();
+
+        // Select the device
+        ECINDAR1 = SPI_CHIP_SELECT;
+
+        // Send erase command
+        ECINDDR = SPI_ERASE_SECTOR_COMMAND;
+        ECINDDR = addr >> 16;
+        ECINDDR = addr >> 8;
+        ECINDDR = addr;
+
+        // Deselect
+        ECINDAR1 = SPI_CHIP_DESELECT;
+        ECINDDR  = 0x00;
+
+        // Wait WIP to be cleared
+        flash_wait();
+
+        flash_exit_follow_mode();
+    }
+}
+
+void flash_enter_follow_mode(void) {
+    // Enter follow mode.
+    ECINDAR3 = SPI_FOLLOW_MODE | SPI_DEVICE;
+    ECINDAR2 = 0xFF;
+    ECINDAR1 = SPI_CHIP_DESELECT;
+    ECINDAR0 = 0x00;
+    ECINDDR = 0x00;
+}
+
+void flash_exit_follow_mode(void) {
+    // Exit follow mode
+    ECINDAR3 = SPI_DEVICE;
+    ECINDAR2 = 0;
+    ECINDAR1 = 0;
+    ECINDAR0 = 0;
+}
+
+void flash_wait(void) {
+    uint8_t status;
+
+    do {
+        // Select the device
+        ECINDAR1 = SPI_CHIP_SELECT;
+
+        // Send command
+        ECINDDR = SPI_READ_STATUS_COMMAND;
+
+        // read status
+        status = ECINDDR;
+
+        // Deselect
+        ECINDAR1 = SPI_CHIP_DESELECT;
+        ECINDDR  = 0x00;
+    } while(status & SPI_STATUS_WIP);
+}
+
+void flash_write_enable(void) {
+    // Select the device
+    ECINDAR1 = SPI_CHIP_SELECT;
+
+    // Send device id command
+    ECINDDR = SPI_WRITE_ENABLE_COMMAND;
+
+    // Deselect
+    ECINDAR1 = SPI_CHIP_DESELECT;
+    ECINDDR  = 0x00;
+}
--- a/src/board/system76/common/flash/wrapper.c
+++ b/src/board/system76/common/flash/wrapper.c
@@ -1,0 +1,98 @@
+/*
+ * Copyright (C) 2020 Evan Lojewski
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#include <stdint.h>
+#include <stddef.h>
+#include <board/flash.h>
+#include <common/ec.h>
+
+#if __EC__ == it5570e
+// IT5570: SCAR0 is stored in processor cache, not in xram
+volatile uint8_t __xdata __at(0x1040) SCAR0L;
+volatile uint8_t __xdata __at(0x1041) SCAR0M;
+volatile uint8_t __xdata __at(0x1042) SCAR0H;
+
+#define SCARL SCAR0L
+#define SCARM SCAR0M
+#define SCARH SCAR0H
+
+#elif __EC__ == it8587e
+// IT8587: SCAR1 is in xram at 0x800-0xC00
+volatile uint8_t __xdata __at(0x1043) SCAR1L;
+volatile uint8_t __xdata __at(0x1044) SCAR1M;
+volatile uint8_t __xdata __at(0x1045) SCAR1H;
+
+#define SCARL SCAR1L
+#define SCARM SCAR1M
+#define SCARH SCAR1H
+
+#else
+#error Unknown EC
+#endif
+
+
+void flash_entry(uint32_t addr, uint8_t * data, uint8_t length, uint8_t command) __reentrant;
+
+// Include flash ROM
+uint8_t __code __at(FLASH_OFFSET) flash_rom[] = {
+    #include <flash.h>
+};
+
+void flash_init(void) {
+    // Use DMA mapping to copy flash data
+    SCARH = 0x80;
+    SCARL = (uint8_t)(FLASH_OFFSET);
+    SCARM = (uint8_t)(FLASH_OFFSET >> 8);
+    SCARH = 0;
+}
+
+void flash_read(uint32_t addr, __xdata uint8_t * data, uint8_t length) {
+    flash_entry(addr, data, length, FLASH_COMMAND_READ);
+}
+
+uint32_t flash_read_u32(uint32_t addr) {
+    uint32_t data;
+
+    flash_entry(addr, (uint8_t *)&data, sizeof(data), FLASH_COMMAND_READ);
+
+    return data;
+}
+
+uint16_t flash_read_u16(uint32_t addr) {
+    uint16_t data;
+
+    flash_entry(addr, (uint8_t *)&data, sizeof(data), FLASH_COMMAND_READ);
+
+    return data;
+}
+
+uint8_t flash_read_u8(uint32_t addr) {
+    uint8_t data;
+
+    flash_entry(addr, &data, sizeof(data), FLASH_COMMAND_READ);
+
+    return data;
+}
+
+void flash_write(uint32_t addr, __xdata uint8_t *data, uint8_t length) {
+    flash_entry(addr, data, length, FLASH_COMMAND_WRITE);
+}
+
+void flash_write_u32(uint32_t addr, uint32_t data) {
+    flash_entry(addr, (uint8_t *)&data, sizeof(data), FLASH_COMMAND_WRITE);
+}
+
+void flash_write_u16(uint32_t addr, uint16_t data) {
+    flash_entry(addr, (uint8_t *)&data, sizeof(data), FLASH_COMMAND_WRITE);
+}
+
+void flash_write_u8(uint32_t addr, uint8_t data) {
+    flash_entry(addr, &data, sizeof(data), FLASH_COMMAND_WRITE);
+}
+
+void flash_erase(uint32_t addr) {
+    flash_entry(addr, NULL, 0, FLASH_COMMAND_ERASE_1K);
+}
--- a/src/common/config.c
+++ b/src/common/config.c
@@ -1,0 +1,295 @@
+/*
+ * Copyright (C) 2020 Evan Lojewski
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#include <board/flash.h>
+#include <common/config.h>
+#include <common/debug.h>
+#include <stddef.h>
+#include <string.h>
+
+// Prevent failures to compile on AVR
+#ifndef __SDCC
+    #define __code
+#endif
+
+config_t *g_config_entries = NULL;
+
+#define CONFIG_ADDRESS  (0x10000L)
+#define CONFIG_SIZE   (1 * 1024)  /* 1K block - code currently assumes this matches the erase size */
+#define CONFIG_PERSISTANT_MAGIC     (0xCF1C0F11)
+#define CONFIG_HEADER_SIZE          (sizeof(uint32_t) + 4) /* Magic number plus four reserved bytes */
+
+#define CONFIG_ENTRIES_START        (CONFIG_ADDRESS + CONFIG_HEADER_SIZE)
+
+uint32_t __code __at(CONFIG_ADDRESS) config_magic;
+uint8_t __code __at(CONFIG_ENTRIES_START) config_persistant[CONFIG_SIZE - CONFIG_HEADER_SIZE];
+
+#define CONFIG_FLASH_ENTRY_EMPTY            (0xFFFFFFFF)
+#define CONFIG_FLASH_ENTRY_CURRENT          (0xFF)
+#define CONFIG_FLASH_ENTRY_OBSOLETE         (0x00)
+
+// config_persistant format
+// (4) bytes: Configuration ID, 0xFFFFFFFF when an entry is not set.
+// (1) bytes: Entry current (valid) or obsolete (invalid)
+// (1) bytes: Entry type (0xFF: 32bit value follows)
+// (4) bytes: Configuration Value
+#define CONFIG_FLASH_ENTRY_ID_OFFSET        (0)
+#define CONFIG_FLASH_ENTRY_CURRENT_OFFSET   (4)
+#define CONFIG_FLASH_ENTRY_TYPE_OFFSET      (5) /* Unused - RFU */
+#define CONFIG_FLASH_ENTRY_VALUE_OFFSET     (6)
+#define CONFIG_FLASH_ENTRY_SIZE             (10)
+
+uint8_t config_validate_magic(void) {
+    uint32_t magic = flash_read_u32(CONFIG_ADDRESS);
+
+    return (CONFIG_PERSISTANT_MAGIC == magic);
+}
+
+uint32_t config_find_free_entry() {
+    if (!config_validate_magic()) {
+        // Write magic.
+        flash_write_u32(CONFIG_ADDRESS, CONFIG_PERSISTANT_MAGIC);
+
+        // Check that that magic wrote properly. If not, we'll need to erase.
+        if (!config_validate_magic()) {
+            return 0;
+        }
+    }
+
+    uint32_t addr = CONFIG_ENTRIES_START;
+    uint32_t limit = addr + sizeof(config_persistant);
+    while(addr < limit) {
+        // configuration present.
+        uint32_t id;
+        id = flash_read_u32(addr + CONFIG_FLASH_ENTRY_ID_OFFSET);
+
+        if(CONFIG_FLASH_ENTRY_EMPTY == id) {
+            // Free entry found.
+            return addr;
+        }
+
+        // Go to the next entry.
+        addr += CONFIG_FLASH_ENTRY_SIZE;
+    }
+
+    return 0;
+}
+
+void config_save_entry_id(uint32_t addr, uint32_t id) {
+    flash_write_u32(addr + CONFIG_FLASH_ENTRY_ID_OFFSET, id);
+}
+
+void config_save_entry_value(uint32_t addr, int32_t value) {
+    flash_write_u32(addr + CONFIG_FLASH_ENTRY_VALUE_OFFSET, value);
+}
+
+bool config_save_entry(config_t *entry) {
+    uint32_t id = (uint32_t)entry->config_id[3] << 24 | (uint32_t)entry->config_id[2] << 16 | (uint32_t)entry->config_id[1] << 8 | entry->config_id[0];
+    int32_t value = entry->value.value;
+    uint32_t addr = config_find_free_entry();
+
+    if (addr && ((addr + CONFIG_FLASH_ENTRY_SIZE - CONFIG_ADDRESS) < sizeof(config_persistant))) {
+        DEBUG("Saving config: %s", entry->config_short);
+        DEBUG(" = %ld\n", value);
+
+        config_save_entry_id(addr, id);
+
+        config_save_entry_value(addr, value);
+
+        return true;
+    } else {
+        return false;
+    }
+}
+
+void config_save_entries() __reentrant {
+    flash_erase(CONFIG_ADDRESS);
+
+    config_t *entry = g_config_entries;
+
+    while (entry) {
+        (void)config_save_entry(entry);
+        entry = entry->next;
+    }
+}
+
+bool config_flash_valid(uint32_t addr) {
+    return (CONFIG_FLASH_ENTRY_CURRENT == flash_read_u8(addr + CONFIG_FLASH_ENTRY_CURRENT_OFFSET));
+}
+
+uint32_t config_flash_id(uint32_t addr) {
+    return flash_read_u32(addr + CONFIG_FLASH_ENTRY_ID_OFFSET);
+}
+
+int32_t config_flash_value(uint32_t addr) {
+    return flash_read_u32(addr + CONFIG_FLASH_ENTRY_VALUE_OFFSET);
+}
+
+uint32_t condif_flash_next(uint32_t addr) {
+    return addr + CONFIG_FLASH_ENTRY_SIZE;
+}
+
+uint32_t config_find_entry(const unsigned char config_id[4]) {
+    uint32_t int_id = (uint32_t)config_id[3] << 24 | (uint32_t)config_id[2] << 16 | (uint32_t)config_id[1] << 8 | config_id[0];
+
+    if(config_validate_magic()) {
+        uint32_t addr = CONFIG_ENTRIES_START;
+        uint32_t limit = addr + sizeof(config_persistant);
+
+        while(addr < limit) {
+            uint32_t id;
+            // configuration present.
+            id = config_flash_id(addr);
+
+            if(id == CONFIG_FLASH_ENTRY_EMPTY) {
+                // no more entries exist.
+                break;
+            }
+
+            if(config_flash_valid(addr)) {
+                if(int_id == id) {
+                    return addr;
+                }
+            }
+
+            // Go to the next entry.
+            addr = condif_flash_next(addr);
+        }
+    }
+
+    // Not found.
+    return 0;
+}
+
+void config_invalidate_entry(uint32_t addr) {
+    if (addr) {
+        flash_write_u8(addr + CONFIG_FLASH_ENTRY_CURRENT_OFFSET, CONFIG_FLASH_ENTRY_OBSOLETE);
+    }
+}
+
+bool config_register(config_t *entry) __reentrant {
+    // Validate entry parameters.
+    if (!entry) {
+        return false;
+    }
+
+    // Short name and desc must be set.
+    if (!entry->config_short || !entry->config_desc) {
+        return false;
+    }
+
+    // Config validated, register it
+    entry->next = g_config_entries;
+    g_config_entries = entry;
+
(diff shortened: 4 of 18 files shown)


src/board/system76/common/kbscan.c
```diff
@@ -7,8 +7,32 @@
...
+    .value = {
+        .min_value = 0,
+        .max_value = 1,
+        .value = 0 /* Default to disable */
+    },
+    .set_callback = NULL,
+};
+
+config_t kbscan_cfg_debounce = {
+    .config_short = "Keyboard Debounce Time",
```
crawfxrd, line 25: This isn't something I would expect to see configurable (I don't expect it to ever change). What value does it provide to be user configurable instead of defined by the firmware?

meklort: I agree that it really isn't useful for most (any?) users to configure the debounce time and can be removed before anything is actually merged.

I'll go ahead and remove it when I next push up changes.

meklort: This is now removed.

meklort:
I've updated the code to handle a full configuration space - the flash block is now erased and the configuration is written back, re-compacted.

With that change, it should be functional without any issues, at least on lemp9.

meklort:
I've added the SPDX identifiers. Note that per the SPDX specification, it's recommended to include the standard license header in addition to the SPDX specified, so I've left it in. If you prefer it removed, I can do so.

I've also added KEYMAP_FN_LOCK to the remaining system76/ boards and made sure everything compiles (I have not tested the Arduino builds).

crawfxrd:
> If you prefer it removed, I can do so.

Yes. SPDX should replace the boilerplate license block with easy to understand identifiers. The "This file is part of..." lines should also be removed (adds no value).

meklort:
As requested, I've removed the standard license header.

src/board/system76/common/flash/main.c
```diff
@@ -0,0 +1,156 @@
...
+#define SPI_STATUS_WIP              (0x01)
+
+void flash_write_enable(void);
+void flash_wait(void);
+void flash_enter_follow_mode(void);
+void flash_exit_follow_mode(void);
+
+// Main flash API entry point. Note: marked as __reentrant so that the main application __data variables are not stomped on.
+void flash_entry(uint32_t addr, uint8_t buffer[], uint8_t size, uint8_t command) __reentrant __critical {
+    if(FLASH_COMMAND_READ == command) {
```
crawfxrd, line 39: fmt: Missing space between control statement and opening brace
nit: Yoda condition

src/board/system76/darp5/keymap/default.c
```diff
@@ -20,3 +20,22 @@ LAYOUT(
...
+),
+LAYOUT(
+    K_ESC, K_F1, K_F2, K_F3, K_F4, K_F5, K_F6, K_F7, K_F8, K_F9, K_F10, K_F11, K_F12, K_PRINT_SCREEN, K_INSERT, K_DEL, K_HOME, K_END, K_PGUP, K_PGDN,
+    K_PLAY_PAUSE, K_1, K_2, K_3, K_4, K_5, K_6, K_7, K_8, K_9, K_0, K_MINUS, K_EQUALS, K_BKSP, K_NUM_LOCK, KT_SCI_EXTRA | SCI_EXTRA_KBD_COLOR, KT_SCI_EXTRA | SCI_EXTRA_KBD_TOGGLE, KT_SCI_EXTRA | SCI_EXTRA_KBD_DOWN,
+    K_TAB, K_Q, K_W, K_E, K_R, K_T, K_Y, K_U, K_I, K_O, K_P, K_BRACE_OPEN, K_BRACE_CLOSE, K_BACKSLASH, K_NUM_7, K_NUM_8, K_NUM_9, KT_SCI_EXTRA | SCI_EXTRA_KBD_UP,
+    K_CAPS, K_A, K_S, K_D, K_F, K_G, K_H, K_J, K_K, K_L, K_SEMICOLON, K_QUOTE, K_ENTER, K_NUM_4, K_NUM_5, K_NUM_6,
+    K_LEFT_SHIFT, K_Z, K_X, K_C, K_V, K_B, K_N, K_M, K_COMMA, K_PERIOD, K_SLASH, K_RIGHT_SHIFT, K_UP, K_NUM_1, K_NUM_2, K_NUM_3, K_NUM_ENTER,
+    K_LEFT_CTRL, KT_FN, K_LEFT_SUPER, K_LEFT_ALT, K_SPACE, K_RIGHT_ALT, K_APP, K_RIGHT_CTRL, K_LEFT, K_DOWN, K_RIGHT, K_NUM_0, K_NUM_PERIOD
+)
+};
```
crawfxrd, line 41: fmt: Missing newline at EOF

src/board/system76/common/flash/main.c
```diff
@@ -0,0 +1,156 @@
...
+#define SPI_READ_STATUS_COMMAND     (0x05)
+
+#define SPI_ERASE_SECTOR_COMMAND    (0xD7)
+
+#define SPI_STATUS_WIP              (0x01)
+
+void flash_write_enable(void);
+void flash_wait(void);
+void flash_enter_follow_mode(void);
+void flash_exit_follow_mode(void);
```
crawfxrd, line 35: Reorder functions before `flash_entry()` to remove the delcarations

src/common/include/common/command.h
```diff
@@ -20,6 +20,16 @@ enum Command {
     CMD_FAN_GET = 7,
     // Set fan speeds
     CMD_FAN_SET = 8,
+
+    // Get Config by name from index.
+    CMD_GET_CONFIG_NAME = 32,
```
crawfxrd, line 25: Why not continue from 9?

crawfxrd:
Testing on a galp3-c:

- When configured for release, system does not boot (doesn't reach first power event; LEDs are off)
- When parport debugging is enabled, system boots, but `system76_ectool config` reports bad values

```
Min: -1 æ÷      æ÷      æ÷ìp!LÀÀÀÀ1ÐÐÐÐ
Max: -11114494
Current Value: -1

: 
Min: 0
Max: 0
Current Value: 2637
```

src/common/include/common/command.h
meklort: The main reason is to allow space so that if additional commands are needed (such as a subgroup, or config type (float / int / string / etc) are needed later on, we don't ahve an issue with mixed command ids.

If a separate command was added and everything was packed, you d then get something like
GET_CONFIG_NAME = 9
OTHER_COMMAND = 10
GET_CONFIG_TYPE = 11

src/board/system76/common/flash/main.c
meklort: fmt: fixed.
Per to the Yoda condition: This is done on purpose. The primary reason is that if the == is accidentally typed as an =, you'll get this error in sdcc:
src/board/system76/common/flash/main.c:39: error 33: Attempt to assign value to a constant variable (=)

If it's the other way around, no error is attributed to that line. That said, sdcc does throw other errors later on:
src/board/system76/common/flash/main.c:53: error 110: conditional flow changed by optimizer: so said EVELYN the modified DOG
src/board/system76/common/flash/main.c:85: error 110: conditional flow changed by optimizer: so said EVELYN the modified DOG
src/board/system76/common/flash/main.c:40: error 126: unreachable code

Basically, on good compilers (clang), it'll throw an error about it and specifically state that if it's intended, you need to wrap it in an additional set of ()'s. On bad compilers, no error will be thrown. On sdcc, no error is thrown but it does cause errors later on to be thrown.

It's part of the coding standards that I usually follow for that specific reason. I'd prefer to leave it as is, but if you want conditionals to be the other way, I can change them.

src/board/system76/darp5/keymap/default.c
meklort: Thanks, I normally have the editor enforce that, but I must have forgotten to set it up (fixed now).

src/board/system76/common/flash/main.c
meklort: So, sdcc doesn't support compiling functions within the same fine in separate sections like gcc/clang based compilers do (text.function_name vs text).
As a result, the function order in the source file is the order that they will appear in the output file. Since the main code assumes flash_entry is at the beginning, I can't just reorder functions in this file.

I can leave it as is, or change this to be something like the following:
(1)
- Split main.c into two files. One will be for the main entry point. This will be placed in the GSINIT0 area to ensure that the file is *always* first in the resulting binary. Note that flash_entry *must* be first in this file. 
- Add a seperate flash_support.c (and flash_support.h file). This is compiled with the default code section. Function prototypes area moved to the .h file, as well as any common command ids, defines, etc.
(2)
Alternatively, I can add a separate crt file in GSINIT0 that jumps to flash_entry.
(3)
Add build system support code to extract the flash function addresses from the flash binary, and pass those to the main firmware image. This would remove the need for flash_entry entirely (each command can be a separate function instead), however it complected the build system a bit. This would also have slightly lower overhead as well.
(4)
Finally, I can leave it as is, just with additional comments clearly stating why flash_entry must be the first function in the file and while main.c must be the first (or only) source file. The main reason I chose this path is because it's easiest to implement.

Let me know your preference.

crawfxrd: Go with 4.

meklort: Ok, comments added / expanded on. Let me know if you want any other changes related to this.

meklort:
> Testing on a galp3-c:
> 
> * When configured for release, system does not boot (doesn't reach first power event; LEDs are off)

Is there anything special that needs to configure for release, or is it the default? I've been using the following:
make BOARD=system76/lemp9
and
make BOARD=system76/lemp9 flash_internal

If there's a different way to test release builds, let me know (-DLEVEL to a value other than 4?). I can try to reproduce it on the lemp9, once I verify that I can debrick with the RT809F.
It'd also be useful to know if this only happens on galp3-c or if it can be reproduced on lemp9 on your side.

> * When parport debugging is enabled, system boots, but `system76_ectool config` reports bad values
> 
> ```
> Min: -1 æ÷      æ÷      æ÷ìp!LÀÀÀÀ1ÐÐÐÐ
> Max: -11114494
> Current Value: -1
> 
> : 
> Min: 0
> Max: 0
> Current Value: 2637
> ```

That's interesting. It implies that the command executed and tried to read the two configuration parameters (fn lock, camera en). This implies that the handshaking (smfi_cmd[1] is ok) is working as expected between the host + ec, just with bad data in smfi_cmd[2] and beyond.

I've also been testing with a modified system76_ectool set to only erase the first 64K, ensuring that my current configuration isn't erased. I'll see if that's a possible reason. (erased flash causing bootup issues vs valid) Lat time I tried there was no issue with an erased config area.

Other possible reasons might be related to the different ITE devices on the two machines. Note that I do not have access to a galp3-c, and so I can't easily test this. If you are able to trace it to a specific commit, that would be useful (all commits should be build-able / run-able.

(this is all of course ignoring the part where the code uses more stack space, and if it was already tight on galp3-c, then it could overflow / cause problems now).

crawfxrd:
> Is there anything special that needs to configure for release, or is it the default?

It's the default. That is, there's no modifications to `board.mk`. As a "debug" build, I enable parport logging.

EC reset hangs at `flash_init()`. Moving it after `gpio_init()` allows reset to complete (system boots), but I don't know what effect that has.

meklort:
> EC reset hangs at `flash_init()`. Moving it after `gpio_init()` allows reset to complete (system boots), but I don't know what effect that has.

Ah, makes sense. I based the flash_init code off of the lemp9 scratch.c file using SCAR0. Looks like galp3-c uses SCAR1 instead. I'll add something to update it to match on the galp3-c

meklort:
I've added a quick change to use SCAR1 instead of SCAR0 for the galp3-c. I don't know if it'll cause any other issue (comment in scratch.c said SCAR1 is in xram at 0x800-0xC00) but feel free to test it. If it works I can either create a separate flash_init function per board, or make the __EC__ switch a bit more generic (move the IT defines to a common include).

Note that with the current code, placing flash_init after gpio_init should not have an impact, and so the fact that it does implies SCAR0 might be doing something I don't expect on the it8587e

(for reference, I specifically use the same configuration as scratch.c for lemp to ensure that I didn't brick anything on lemp9. Code was tested conditionally as well originally such that the flash access was only triggered via an smfi command).

crawfxrd:
`system76_ectool config` when using `SCAR1` works:
```
Default Camera State: When set, the camera is enabled on system reset
Min: 1
Max: 0
Current Value: 1

Fn Lock: When enabled, F1-F12 will activate their default function
Min: 1
Max: 0
Current Value: 0
```

src/board/system76/common/flash/main.c
crawfxrd: The convention used throughout the code is to have constants on the right hand side. We don't enforce any style (yet), but I would prefer to keep it consistent with existing code.

src/common/config.c
```diff
@@ -0,0 +1,314 @@
...
+#include <string.h>
+
+#define CONFIG_FLASH_SIZE       (4 * 1024)      /* 4K sector to allow easy erasing */
+
+// Prevent failures to compile on AVR
+#ifndef __SDCC
+    #define __code
+#endif
+
+config_t *gConfigEntries = NULL;
```
crawfxrd, line 20: Use snake case

src/common/config.c
```diff
@@ -0,0 +1,314 @@
...
+    entry->next = gConfigEntries;
+    gConfigEntries = entry;
+
+    // Read the stored value from flash, if present.
+    uint32_t flash_addr = config_find_entry(entry->config_short);
+    if (flash_addr) {
+        int32_t saved_value = config_flash_value(flash_addr);
+
+        volatile int64_t min = entry->value.min_value;
+        volatile int64_t max = entry->value.max_value;
```
crawfxrd, line 213: Unnecessary volatile?

crawfxrd:
Suppose the format changes to handle a new config. How will the EC and tool detect and handle this?

I've previously used a simple incrementing version number to detect such a case. It doesn't need to be implemented now, but depending on ideas on how this will be handled (if at all), we may want to reserve some some bytes at the start of the config space.

src/common/config.c
meklort: I was seeing this improve DSEG/OSEG usage here as well. I've removed the volatile keywords the memory issue's improved in other ways since I ran into the problem.

src/common/config.c
meklort: should be fixed now.

src/common/config.c
```diff
@@ -0,0 +1,314 @@
...
+        uint8_t len;
+        len = flash_read_u8(addr + CONFIG_FLASH_ENTRY_LEN_OFFSET);
+
+        if(CONFIG_FLASH_ENTRY_LENGTH_INVALID == len) {
+            // Free entry found.
+            return addr;
+        }
+
+        // Go to the next entry.
+        addr = addr + CONFIG_FLASH_ENTRY_NAME_OFFSET + len;
```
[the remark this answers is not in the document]
meklort: With SDCC, this actually results in two different binaries being produced. From my initial testing, "addr = addr +" appeared to alleviate some memory issues I was running into (since improves with the i2c change).. 

Form my retesting today, there's actually a 5 byte hit in code size when using addr = addr + instead of +=.
I've updated this.

src/board/system76/common/kbscan.c
```diff
@@ -7,8 +7,20 @@
 #include <board/keymap.h>
 #include <board/pmc.h>
 #include <board/power.h>
+#include <common/config.h>
 #include <common/debug.h>
 
+config_t kbscan_cfg_fnlock = {
+    .config_short = "Fn Lock",
+    .config_desc = "When enabled, F1-F12 will activate their default function",
```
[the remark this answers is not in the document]
meklort: Sure, I can see that. I've changed it to say alternative.

meklort:
> Suppose the format changes to handle a new config. How will the EC and tool detect and handle this?
> 
> I've previously used a simple incrementing version number to detect such a case. It doesn't need to be implemented now, but depending on ideas on how this will be handled (if at all), we may want to reserve some some bytes at the start of the config space.

One option could be using a different magic number to differentiate between the two.

I can also see where it might be good to add a "config type" byte in the config entry itself. this could be placed in the entry header itself. On older version of the firmware, unknown types can by skipped as the size is known. 

Also note that currently the code probably won't handle downgrading the most gracefully - it'd effectively try to overwrite the magic number w/o erasing, silently fail, and then try to use the following data anyway. I'll clean it up a bit and add something to future proof it.

Do note that I'll be offline for about a week, so you won't see any updates from me for a bit.

src/common/include/common/config.h
```diff
@@ -0,0 +1,63 @@
...
+
+typedef struct {
+    int32_t max_value;
+    int32_t min_value;
+    int32_t value;
+} config_value_t;
+
+struct config_t {
+    const char *config_short;
+    const char *config_desc;
```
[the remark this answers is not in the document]
meklort: Would you please elaborate on why having strings concerns you? 

At a minimum, I think the human readable name (config_short) should be here. I don't see a downside for having both really - there's currently an extra 96KB of flash that isn't in use. The string copy code is also limited to the buffer size, 256 bytes, and so there is not really a risk of overrunning memory. That said, the config save code is not limited in the same way - it's only limited to the side of the config space. I can go ahead and limit this further, however.

Personally, I'm a lot more concerned about the ability to erase and write to flash. To me it seems more likely that the code will accidentally brick itself. One option to make this safer is to explicitly check bounds in the flash application.

The reason I did it this way is to enable easy autoconfig and so multiple tools don't need to be kept in sync. Possible host-side tools include the ectool, possibly a settings plugin for gnome, or a UEFI application. With the way the code is now, valid settings can be read from the firmware and customized automatically. If the name or description were moved, you now need to both update the host tool and the firmware at the same time. For a gnome setting plugin, this isn't really supported right now. If the ectool is automatically installed in Pop!, Then this also wouldn't be ensured to be in sync. for a UEFI application, you could enforce this, howerver.

One other benifit is that it's very easy to find the setting in a config dump when manually inspecting.

All that said, moving just the description would still allow for the human readable name (just no description).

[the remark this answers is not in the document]
crawfxrd: > Would you please elaborate on why having strings concerns you?

I'm mainly concerned with string lengths: the space they'll take up and reading from/writing to flash. I was also thinking about changing these strings (requiring a firmware update) and translating these strings (but I suppose gettext would work for user-space applications?).

For example, the setting could be represented by a single byte. This would mean only reading/writing 1 byte instead of _n_ bytes for the identifier. Each entry would then also be the same size. Conceptually, this is much simpler to iterate and find settings (no need for `flash_strncmp()` or tracking lengths).

(I can be convinced either way, though. I've only worked with firmware that has settings defined at compile time, so I don't have much exposure with this stuff. Even the Linux kernel embeds the name and description into modules.)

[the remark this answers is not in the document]
meklort: I agree that the search isn't the most optimal, and it takes up more space. I guess it comes down to priorities. I tend to lean towards autoconfig whenever possible. (I also don't have variable length strings within the firmware I normally work on, although most of the firmware I work on is configurable by the user in various other ways).

You also have a good point about translations, I hadn't thought about that. You'd still need to know the strings (or an I'd) in there user app to have a translation.

I could be convinced to have, say, a 4 byte id such as 'KFNL' for keyboard function lock and 'CCDE' for camera enable, for instance. (Maybe even just 16bits is enough). This could either replace or supliment the human readable name/desc while still having some meaning.

The short Id (1 to 4 bytes) would be used for, say, searches and storing in flash.

I can update everything in this pull request to use the smaller 1-4 byte id for searches / commands. I propose leaving the desc / name so that at least English translations can be discovered, but if you'd prefer I can (reluctantly) move those to the ectool.

If you think it's worth doing, as a second step, I could also see adding a resource table in flash (longer descriptions, translations, keymaps, other could be stored here). This could then completely replace the larger strings.
The resource table could be stored in end of flash (outside of the 64kb region) and read by a command. 

I can add another pull request adding a resource table. I'd then add an command to allow the host to read out resource, and move the English translations to said table. As a longer term feature, it might be nice to allow keymaps to be set here, not just read. A user could generate them using something like the qmk configurator.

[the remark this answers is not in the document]
crawfxrd: > The short Id (1 to 4 bytes) would be used for, say, searches and storing in flash.

That's a good compromise. Keep the strings, and add an identifier for enumeration/lookup. I'd be fine with either 1 byte (binary values) or 4 bytes (ASCII short-code). 2 bytes would be too much for potential config identifiers\* and too short for an ASCII identifier.

\* Unless a change in semantics for one results in a new identifier, similar to adding new commands while maintaining compatibility with old ones.

(Using a 2 byte value like that sounds like it covers my concerns about changing the format of a config in the future.)

I don't think the resource table is worth doing; Descriptions and translations can be handled by higher level tools and documentation (our BIOS firmware is also English-only). The keymap idea is intriguing, though, since everyone wants something slightly different.

crawfxrd:
Going over the drawbacks section:

> * This code calls between two applications. The flash API is a separate image which has no knowledge of XSEG / DSEG usage. As a result, the a developer modifying the must take special not to mangle existing registers.

This one will require documentation on. Maybe a block comment at the top of one of the files, or a section in a new `doc/` file on this feature. (This definitely sounds like something I will screw up.)

> * Any mistakes in the flash writing code could accidentally erase or overwrite the main EC binary, bricking the device. Special care need to be taken to ensure that a device is not bricked. Note: Additional checking could be added to the flash API to limit writes to a small window.

What checks would these be? It looks like the checks in place prevent getting an address outside the config space. But it's probably reasonable to add them.

> * The configuration APIs only support a single 32bit value. In some cases, a larger value may be desired. As an example, the keyboard layout could be stored in flash using a similar mechanism. This would allow custom keymaps. At present, this use case is not supported.

This sounds like a good "version 2" project.

> * The code currently does a linear search in flash. As the ~4KB~ 1KB configuration space fills up, read/write access time can be impacted. This can slow down the initial boot, as well as reconfiguration time by the user.

I think this can be punted. A linear search seems sufficient for now.

But thinking of what we do for coreboot's smmstore [\[1\]](https://github.com/system76/firmware-smmstore/blob/c70661a4e3d6a2413313a75b8deb62fccd275e62/src/smmstore.rs#L69): A "compact" command could be implemented to erase the space and write back the valid entries.

meklort:
> Going over the drawbacks section:
> 
> > * This code calls between two applications. The flash API is a separate image which has no knowledge of XSEG / DSEG usage. As a result, the a developer modifying the must take special not to mangle existing registers.
> 
> This one will require documentation on. Maybe a block comment at the top of one of the files, or a section in a new `doc/` file on this feature. (This definitely sounds like something I will screw up.)
> 

I think part of the main PR comments could be added to a new docs file. I'll look into doing that in addition to comments as needed in the flash code.

> > * Any mistakes in the flash writing code could accidentally erase or overwrite the main EC binary, bricking the device. Special care need to be taken to ensure that a device is not bricked. Note: Additional checking could be added to the flash API to limit writes to a small window.
> 
> What checks would these be? It looks like the checks in place prevent getting an address outside the config space. But it's probably reasonable to add them.
> 

My main thought would be to add checks for the config space in the flash main.c file itself . The thought would be that if there's a bug due to communication with the two different binaries, the EC would still be bootable as the only thing that could be erased / written is the config space. The other thought is it'd add a second layer of checks in case if there's a bug in the config code. It's not strictly necessary, but I want to be extra careful when doing anything that has to do with erasing or writing to flash that's shared with the main executable to mitigate risks.

> > * The configuration APIs only support a single 32bit value. In some cases, a larger value may be desired. As an example, the keyboard layout could be stored in flash using a similar mechanism. This would allow custom keymaps. At present, this use case is not supported.
> 
> This sounds like a good "version 2" project.
> 

Agreed, that doesn't need to be included in this pull request.

> > * The code currently does a linear search in flash. As the ~4KB~ 1KB configuration space fills up, read/write access time can be impacted. This can slow down the initial boot, as well as reconfiguration time by the user.
> 
> I think this can be punted. A linear search seems sufficient for now.
> 

Agreed, The flash size is small enough that it shouldn't be noticeable.

> But thinking of what we do for coreboot's smmstore [[1]](https://github.com/system76/firmware-smmstore/blob/c70661a4e3d6a2413313a75b8deb62fccd275e62/src/smmstore.rs#L69): A "compact" command could be implemented to erase the space and write back the valid entries.

This actually automatically happens right now any time the flash fills up - it compacts. The main reason is to limit flash writes (and minimizes slowdowns due to erasing), but as a side effect I can easily add a compact command that could when needed. I should probably add a check to ignore the command if it's already as compact as possible, though.

src/common/include/common/config.h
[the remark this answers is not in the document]
meklort: The code now uses a 4 byte ascii identifier.

Note that in SDCC setting a uint32_t value to  'CCDE'  actually only sets it to 'C', unlike for normal 32bit targeting compilers. As such I have it using {'C', 'C', 'D', 'E'} instead even though I'd prefer not to.

crawfxrd:
LGTM. I think this is nearly done. @jackpot51 this seems like a good point to get some feedback from you.

meklort:
I've rebased this and added the compact command as requested earlier.

crawfxrd:
Keymap changes need to be added for the oryp6.

This commit looks unneeded: 744b85b6fb235 ("common: Enable preprocessor #ifs based on \_\_EC\_\_")

meklort:
> Keymap changes need to be added for the oryp6.

Done. I've also added changes for the new keymaps as well.
 
> This commit looks unneeded: [744b85b](https://github.com/system76/ec/commit/744b85b6fb23585166a7af71cfc538ce1b8eaadb) ("common: Enable preprocessor #ifs based on __EC__")

This is needed due for the compiler defines based on \_\_EC__ to work properly in 215343f45872465476c208739e45227ced6610c0

Previously, \_\_EC__ was defined to an undefined value. As a result, any preprocessor conditionals based on \_\_EC__ could not be evaluated (always evaluated as false). Commit  744b85b allows the preprocessor to evaluate \_\_EC__ to a valid value, and so conditionals now evaluate properly.

crawfxrd (dismissed):

meklort:
I've removed the two commits that enable configuring the function lock and CCD_EN default state as requested. Those are now  in the separate pull request #80 
(I think this is what you wanted, but please let me know if it's not the case.)

src/board/system76/common/flash/wrapper.c
```diff
@@ -0,0 +1,98 @@
...
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#include <stdint.h>
+#include <stddef.h>
+#include <board/flash.h>
+#include <common/ec.h>
+
+#if __EC__ == it5570e
+// IT5502: SCAR0 is stored in processor cache, not in xram
```
jackpot51, line 13: This should be IT5570

src/board/system76/common/flash/main.c
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * NOTE: This *must* be the first function in this file to ensure that it is placed
+ *          first in the resulting binary. This is required to ensure that address
+ *          matches the address (FLASH_OFFSET) for flash_entry in wrapper.c.
+ * NOTE: __reentrant so that parameters and temperary vairables are placed on the
+ *          stack, ensuring the main application __data variables are not stomped on.
+ * NOTE: __critical to ensure interrupts are disabled. This does mean that interrupt
+ *          such as the timer will be block until flash acccess is complete
+ */
+void flash_entry(uint32_t addr, uint8_t * data, uint8_t length, uint8_t command) __reentrant __critical {
```
jackpot51, line 48: If we don't expect to use this for flashing the code space below 64 KiB, perhaps doing this would be good:

```
if (addr < 0x10000 || addr >= 0x20000) return;
```

src/board/system76/common/flash/wrapper.c
```diff
@@ -0,0 +1,98 @@
...
+volatile uint8_t __xdata __at(0x1040) SCAR0L;
```
jackpot51, line 14: Consider putting these into src/ec instead.

src/board/system76/common/flash/wrapper.c
```diff
@@ -0,0 +1,98 @@
...
+#define SCARH SCAR1H
+
+#else
+#error Unknown EC
+#endif
+
+
+void flash_entry(uint32_t addr, uint8_t * data, uint8_t length, uint8_t command) __reentrant;
+
+// Include scratch ROM
```
jackpot51, line 39: Include flash ROM

src/board/system76/common/flash/wrapper.c
```diff
@@ -0,0 +1,98 @@
...
+#define SCARL SCAR1L
+#define SCARM SCAR1M
+#define SCARH SCAR1H
+
+#else
+#error Unknown EC
+#endif
+
+
+void flash_entry(uint32_t addr, uint8_t * data, uint8_t length, uint8_t command) __reentrant;
```
jackpot51, line 37: If you have a .h file shared between flash ROM and the main ROM, and included it in flash/main.c, then this definition would be checked by the compiler to match the definition in flash/main.c

src/board/system76/common/flash/flash.mk
```diff
@@ -0,0 +1,56 @@
...
+
+# Add minimal source from other directories
+FLASH_SRC+=
+
+FLASH_BUILD=$(BUILD)/flash
+FLASH_OBJ=$(patsubst src/%.c,$(FLASH_BUILD)/%.rel,$(FLASH_SRC))
+FLASH_CC=\
+	sdcc \
+	-mmcs51 \
+	--model-large \
```
jackpot51, line 25: The large model will potentially use data space also used by the main ROM. Is there a reason to not use --model-small like the scratch ROM?

src/common/include/common/ec.h
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+#ifndef _COMMON_EC_H
+#define _COMMON_EC_H
+
+#ifndef __EC__
+#error __EC__ must be defined
+#endif
+
+#define it5570e 5570
```
jackpot51, line 12: Is this required? The scratch ROM uses the correct SCAR registers depending on __EC__ definition without having these

tool/src/main.rs
```diff
@@ -296,11 +296,105 @@ unsafe fn fan_set(index: u8, duty: u8) -> Result<(), Error> {
...
+
+    ec.config_compact()
+}
+
 fn usage() {
     eprintln!("  console");
     eprintln!("  flash [file]");
     eprintln!("  flash_backup [file]");
     eprintln!("  fan [index] <duty>");
+    eprintln!("  config [name] [value]");
```
jackpot51, line 396: If optional arguments are in [], then flash, flash_backup, fan, and print need to be adjusted.

src/common/include/common/config.h
```diff
@@ -0,0 +1,72 @@
...
+typedef void (*config_callback_t)(struct config_t *entry);
+
+typedef struct {
+    int32_t max_value;
+    int32_t min_value;
+    int32_t value;
+} config_value_t;
+
+struct config_t {
+    const unsigned char config_id[4]; /* Note: SDCC does not support uint32_t being set to values such as 'ASCI', as such, this an explicit array instead */
```
jackpot51, line 26: Please describe the purpose of each field here in comments

src/common/config.c
```diff
@@ -0,0 +1,295 @@
...
+
+        config_save_entry_value(addr, value);
+
+        return true;
+    } else {
+        return false;
+    }
+}
+
+void config_save_entries() __reentrant {
```
jackpot51, line 108: What happens in the case that the config entries even when compacted overflow available flash space?

src/common/include/common/ec.h
meklort: Yes, this is needed. The code in scratch.c actually doesn't work how you think.
As a quick example, I modified scratch.c as follows:
<code> 
#if \_\_EC__ == it5570e
</code>
<code>
+#error it5570e
</code>
and
<code>
 #elif \_\_EC__ == it8587e
</code>
<code>
+#error it8587e
</code>

When building galp3-c, the following error is reported:
<code>
sdcc -mmcs51 --model-large --code-size 65536 --Werror -Isrc/common/include -D__FIRMWARE_VERSION__=2020-08-02_2e6487d-dirty -Isrc/board/system76/galp3-c/include -D__BOARD__=system76/galp3-c -DI2C_SMBUS=I2C_0 -DPS2_TOUCHPAD=PS2_3 -DCHARGER_CHARGE_CURRENT=1536 -DCHARGER_CHARGE_VOLTAGE=13056 -DCHARGER_INPUT_CURRENT=1920 -DLEVEL=4 -Isrc/board/system76/common/include -DSCRATCH_OFFSET=1024 -DSCRATCH_SIZE=1024 -Ibuild/system76/galp3-c/2020-08-02_2e6487d-dirty/include -DFLASH_OFFSET=2048 -DFLASH_SIZE=1024 -Ibuild/system76/galp3-c/2020-08-02_2e6487d-dirty/include -Isrc/ec/it8587e/include -D__EC__=it8587e -Isrc/arch/8051/include -D__ARCH__=8051 -o build/system76/galp3-c/2020-08-02_2e6487d-dirty/board/system76/common/scratch.rel -c src/board/system76/common/scratch.c
src/board/system76/common/scratch.c:15:2: error: #error it5570e
</code>
You'll node that  #error it5570e is reported (we are in the 5570e case), even though \_\_EC__ is defined as 8587.

The reason this is happening is because the compiler is that both it8587e and it5570e are not defiend - that is, they have the same value. As a result, only the first case ever happens in scratch.c.

src/board/system76/common/flash/main.c
meklort: Adding a check like this sounds good to me. If we ever want it to write to the lower 64KB, the code can always change at that point.
I'll also look at making sure the length doesn't cause it to overflow.

src/common/include/common/ec.h
meklort: Also note that adding the new include in causes the expected error to be printed:
#include <common/ec.h>
src/board/system76/common/scratch.c:25:2: error: #error it8587e

src/board/system76/common/flash/flash.mk
meklort: From the sdcc manual:
When the small model is used variables without an explicitly specified intrinsic named address space are allocated in the internal ram.
When the medium, large or huge model is used all variables declared without specifying an intrinsic named
address space will be allocated into the external ram, this includes all parameters and local variables (for nonreentrant functions). Medium model uses pdata and large and huge models use xdata.

Effectively, -small causes the main data segment to be mangled, while large causes external data to be mangled.
In reality, it shouldn't matter so long as reentrant is used, and this ensures that only the stack is used.

Unfortunately, the main reason for one over the other is due to the runtime libraries that get pulled in. When -small is used, _gptrput gets pulled in with PARAM_2 in OSEG (data). When using -large, this is placed in XSEG instead (xdata). Note that ideally this would be reentrant as well, however the sdcc doesn't have a reentrant version of _gptrput.

The xdata location that is allocated for _gptrput is unused according to ec.map and flash.map, while the location in data was already used.

tool/src/main.rs
meklort: Looks like I misread this originally. I've fixed it to match fan with <> for optional arguments.

src/board/system76/common/flash/wrapper.c
meklort: This should be fixed now.

src/board/system76/common/flash/main.c
meklort: I've added this check with an additional check for addr + length.
Note that both addr and addr + length are checked to ensure there's no wrapping. (MAX_UINT start address wrapping to 0)

src/board/system76/common/flash/wrapper.c
meklort: Done

src/common/config.c
meklort: If the compacted area overflows config space, the end result is that some of the values will not be saved.

All entries up until the config space is full will be saved. Once it's full, however, config_save_entry will return false and config_save_entries will silently fail.

Note that in this case, There's a fundamental issue in that more configuration parameters exist than are possible to store, and the only possible solutions are to either (1) reduce the number of config entries or (2) increase the config space from 1KB to larger.

I can update this code to have config_save_entries return a bool and propagate that up through to the host config_compact command. During regression testing for release, it'd then be a good idea to always issue a config_compact command and check for success to ensure that the firmware doesn't have this problem.

meklort: The error should now be propagated to the host both when setting a config value (if compacting is needed and it fails) and if compacting is manually requested.

rellimmot:
Is this a dead effort? I'm wondering if y'all ever plan on fielding this feature or if I should seek alternate means of Fn Lock/Invert.

leviport:
@rellimmot Thank you for so thoroughly expressing interest in this feature. If the PR is open, the effort is not dead.

crawfxrd:
Due to the limitation of only supporting u32 and the desire to move away from modifying EC flash at runtime, I will be closing this issue.

If modifying flash space is required (such as for persistent battery charging thresholds), I would prefer to iterate on #256.

I have started https://github.com/system76/firmware-open/issues/571 for moving policies to host firmware.

PR #67, opened 2020-05-27
Addw2 support

jackpot51:
Part of https://github.com/system76/firmware-open/pull/84


src/board/system76/common/main.c
```diff
@@ -74,17 +74,17 @@ void main(void) {
 #endif
 
     // Allow CPU to boot
-    gpio_set(&SB_KBCRST_N, true);
+    //gpio_set(&SB_KBCRST_N, true);
```
crawfxrd, line 77: Does the EC not have to do all these, or are these disabled for debug/development purposes?

jackpot51:
Tested darp6 and lemp9 successfully - after the last few commits

src/board/system76/common/gctrl.c
```diff
@@ -2,5 +2,8 @@
 
 void gctrl_init(void) {
     SPCTRL1 = 0x03;
+#ifdef POST_DEBUG
+    SPCTRL1 |= 0xC8;
```
crawfxrd, line 6: This looks specific to 5570.

src/board/system76/common/main.c
```diff
@@ -142,5 +129,15 @@ void main(void) {
         smfi_event();
         // Idle until next timer interrupt
         //Disabled until interrupts used: PCON |= 1;
+
+    #ifdef POST_DEBUG
+        // Read POST codes
+        if (P80H81HS & 1) {
```
crawfxrd, line 135: This looks specific to 5570.

jackpot51: Yeah, that's correct. I can remove it

PR #68, opened 2020-05-29
Add a keymap putting Insert on the PrtSc key

arodland:
The lemp9 doesn't have an Insert key, but PrtSc is right there
next to Delete, and while I like taking screenshots, it really
doesn't deserve to be the primary function of a key.

Add an alternative keymap that makes the PrtSc key act as Insert,
leaving the print screen function on Fn+PrtSc

build with "make BOARD=system76/lemp9 KEYMAP=ins-prtsc"


arodland:
I'm running this personally... just pushing it up here in hopes that others might find it useful, or even that it would find its way in as a configure toggle in #63.

PR #70, opened 2020-06-08
gaze15 and oryp6

jackpot51:
Still need to add the 1650 variant, and see if a split between the two is necessary


cognifloyd:
Wow. I can't tell you how excited I am to see gaze15 ec. BIOS and firmware has always seemed like "abandonware" to me. I can't wait to get my hands on a machine that I might be able to tweak.

Sometimes the tweaks I want to do, like changing the Caps lock key's function or reorganizing the meta keys (super/ctrl/alt/fn/etc) has been beyond anything os-level tools (xmodmap, kde keyboard layouts) can handle. the possibility of tweaking this at the EC level... awe inspiring!

I wish I had more than just encouragement to offer in supporting what you're doing. As soon as I can, I'll buy a system76 laptop. open-firmware + open ec clinches it. :smile:

Issue #74, opened 2020-06-23
Disable key repeat on certain hotkeys


leviport:
On lemp9 and oryp6 I noticed that holding Fn+F2 makes the display strobe on and off. We should probably disable key repeating on that hotkey, as well as some others.

The only ones where I think we would want it are:
* Volume up and down: Fn + F5 and F6
* Display brightness up and down: Fn + F8 and F9
* Keyboard brightness up/down (for the machines that have up and down for keyboard brightness, at least)

Lemp9's keyboard light level hotkey (Fn+F4) should maybe also have repeating turned off as well. It looks cool, but doesn't seem useful in practice.

crawfxrd:
Disabling repeat won't matter for normal keys because GNOME implements its own key repeat to control these (Universal Access > Typing > Repeat Keys).

PR #78, opened 2020-07-08
kbscan: Disable key repeat for certain SCI keys

crawfxrd:
Resolves: #74

### DIFF
--- a/src/board/system76/common/kbscan.c
+++ b/src/board/system76/common/kbscan.c
@@ -215,6 +215,22 @@ bool kbscan_press(uint16_t key, bool pressed, uint8_t * layer) {
     return true;
 }
 
+static inline bool key_should_repeat(uint16_t key) {
+    switch (key) {
+    case K_TOUCHPAD:
+    case (KT_SCI | SCI_AIRPLANE_MODE):
+    case (KT_SCI | SCI_CAMERA_TOGGLE):
+    case (KT_SCI | SCI_DISPLAY_TOGGLE):
+    case (KT_SCI | SCI_SUSPEND):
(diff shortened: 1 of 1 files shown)


crawfxrd:
Proprietary firmware on oryp5 has key repeat for display backlight level, but not keyboard backlight level. May want to explicitly list all keys to prevent repeat on instead of all `KT_SCI` keys.

src/board/system76/common/kbscan.c
```diff
@@ -215,6 +215,21 @@ bool kbscan_press(uint16_t key, bool pressed, uint8_t * layer) {
     return true;
 }
 
+static inline bool key_should_repeat(uint16_t key) {
+    switch (key) {
```
jackpot51, line 219: Also need touchpad toggle

crawfxrd:
Holding the display backlight toggle key while in firmware-setup still flickers the display on and off.

jacobgkau:
Tested this:
- Volume up/down still have key repeat on; it respects GNOME's key repeat setting, so that's fine.
- Display brightness up/down still have key repeat on; they do not respect GNOME's key repeat setting.
- Keyboard backlight brightness up/down still have key repeat on; they do not respect GNOME's key repeat setting.
- All other keys have repeats turned off (within the OS; like Tim noted, within firmware-setup the display on/off repeats very quickly.)

This is an improvement since it stops the display on/off and keyboard backlight on/off from flickering. Are we intending to make the keyboard brightness and display brightness no longer repeat, or are we intending to leave those as-is? (Also wasn't sure if more work would be done to stop the flickering in firmware-setup.)

jackpot51:
I think we want brightness keys to repeat

PR #80, opened 2020-07-22
Config values: Enable Configurable CCD_EN default state and keyboard function lock.

meklort:
This pull request is based on #63 as requested and includes additional commits to enable a configurable keyboard function lock as well as CCD_EN default states.


meklort:
Note that I'm not sure how to, or if it's possible to mark this as based off of #63 however the diff can be seen here:
https://github.com/meklort/ec/compare/config-pull...meklort:config-values

crawfxrd:
Ref: https://github.com/system76/ec/pull/63#issuecomment-2249249643

PR #82, opened 2020-07-23
docs: Add I2C pins for more models

crawfxrd:
Add battery pins used for I2C debugging on additional models. Change pin names to those used in laptop  service manuals.

### DIFF
--- a/doc/debugging.md
+++ b/doc/debugging.md
@@ -68,18 +68,31 @@ unsupported board, may result in damaged board components.**
 
 ### Wiring the target
 
-These steps apply to the following models:
-- darp5
-- darp6
-- galp3-c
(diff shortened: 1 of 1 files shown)


doc/debugging.md
```diff
@@ -105,8 +118,8 @@ sudo tio -b 115200 -m INLCRNL /dev/ttyACM0
 4. Ground target to host
     - Connect USB cable from USB-C port on target to host
 5. Connect target to host
```
jackpot51, line 120: They will also need to flash firmware to the trinket which takes I2C writes to 0x76 and prints them to the console

jackpot51: This can be done in another commit

PR #86, opened 2020-08-07
Docs update 

crawfxrd:
- Add a more comprehensive document describing the Mega 2560 and FPC breakout board
- Move flashing information to a separate file
- Point users to firmware-open for installing full set of dependencies


jackpot51:
I would recommend moving the documentation section up higher in the readme. Perhaps move the features lists to their own document

PR #87, opened 2020-08-13
battery: Implement charging thresholds

crawfxrd:
Implement battery charging thresholds based on the [sysfs `charge_control_{start,end}_threshold` properties](https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-class-power).

Battery Charging Start Threshold:
- Valid values are [0, 100]
- A value of 0 disables the feature, leaving the charger enabled on AC

Battery Charging End Threshold:
- Valid values are [0,100]
- A value of 100 disabled the feature, leaving the charger enabled on AC

The stop threshold must be configured for the start threshold to work.

_Note_: This implementation is only configurable at build time. #63 and/or ACPI methods can be added later to allow configuring at runtime.

Ref: system76/coreboot#15
Ref: torvalds/linux@fa93854f7a7ed63d054405bf3779247d5300edd3


crawfxrd:
> When not charging because the charge is above the lower threshold or at the upper threshold, the "time remaining" still shows an amount of time (or sometimes just "Estimating...") even though it's not actually charging.

This seems to be an [old issue](https://bugs.launchpad.net/ubuntu/+source/indicator-power/+bug/1231550) with gsd-power/upower.

> The charge sometimes stops at one percentage below the upper threshold, e.g. 69% instead of 70% or 79% instead of 80%. (I'm assuming it's barely reaching the upper threshold and then immediately falling back down one percentage when it stops charging, and it won't start charging again because it's still above the lower threshold.)

Correct. It is currently comparing the values using `>=`/`<=`, so it will start/stop charging as soon as the battery data is updated and has those values. (The EC updates battery data every ~1s.) I think this is the behavior in ThinkPads, but I'm not entirely sure.

(I believe this is more precise that what the sysfs documentation implies, which states "above/below which", meaning it would starts/stops at -1/+1 of whatever the value is.)

ids1024:
> Valid values are [0,100]

Hm. Does this mean someone could do something silly, like configure their battery to only charge between 0 and 1 percent? Is there any way that could cause serious issues, or should it not really be a problem because they can always boot fine on AC power to change the setting?

I'm not sure what range of valid values other manufacturers use.

crawfxrd:
> Does this mean someone could do something silly, like configure their battery to only charge between 0 and 1 percent?

Yes.

>  Is there any way that could cause serious issues

The worst things it can do are more rapidly degrade the life of the battery, or prevent booting on battery power.

> they can always boot fine on AC power to change the setting?

Correct

> I'm not sure what range of valid values other manufacturers use.

The [ASUS](https://github.com/torvalds/linux/blob/15bc20c6af4ceee97a1f90b43c0e386643c071b4/drivers/platform/x86/asus-wmi.c#L410), [Huawei](https://github.com/torvalds/linux/blob/15bc20c6af4ceee97a1f90b43c0e386643c071b4/drivers/platform/x86/huawei-wmi.c#L347), and [ThinkPad](https://github.com/torvalds/linux/blob/15bc20c6af4ceee97a1f90b43c0e386643c071b4/drivers/platform/x86/thinkpad_acpi.c#L9557) Linux drivers all limit to this range. (ThinkPad limits it to [0,99] and [1,100], since a value of 100 for start or 0 for end are not possible/don't make sense.)

jackpot51 (approved):

PR #98, opened 2020-09-17
Add SPDX license identifiers

crawfxrd:
Add license information to files. Mostly automated with:

    find src/ -name '*.[c,h]' | xargs sed -i '1s,^,// SPDX-License-Identifier: GPL-3.0-only\n\n,'
    find src/ -name '*.mk' | xargs sed -i '1s,^,# SPDX-License-Identifier: GPL-3.0-only\n\n,'


jackpot51:
Any reason .c and .h files have a different comment format? I'd probably prefer the one used in .c files

crawfxrd:
It's the style used by the Linux kernel: https://www.kernel.org/doc/html/latest/process/license-rules.html#license-identifier-syntax

Given it's done only because of limitations in old tools that don't affect use, I can convert them.

jacobgkau (approved):
All of the changes are comments. Confirmed I am still able to build and flash a machine.

Issue #102, opened 2020-09-21
Do not wake on keypress with lid closed


jackpot51:
This has woken up some customer's machines while they are in bags

PR #106, opened 2020-09-21
kbscan: Prevent wake when lid is closed

crawfxrd:
Only allow key presses to wake from S3 when the lid is open.

Resolves: #102


jackpot51 (approved):

leviport (approved):
lemp9 no longer wakes when keys on the internal keyboard are pressed while the lid is closed. As expected, it still wakes on keypress from internal keyboard when lid is open, and still wakes on keypress on USB keyboard while lid is closed and an external display is attached.

PR #110, opened 2020-09-25
Reset EC config if Fn+Esc held during power on

crawfxrd:
Reset EC configuration; Currently only battery charging thresholds.

To test:

- Plug into AC power
- Flash this firmware
- Modify charging thresholds
- Power off and on
- Confirm charging thresholds are unchanged
- Power off
- Hold Fn+Esc and press power button to power on
  - Once the power LED turns green, they can be released
- Confirm charging thresholds have reset to 0 and 100

### DIFF
--- a/src/board/system76/addw2/include/board/keymap.h
+++ b/src/board/system76/addw2/include/board/keymap.h
@@ -46,4 +46,12 @@
 // Keymap
 extern uint16_t __xdata KEYMAP[KM_LAY][KM_OUT][KM_IN];
 
+// Position of physical Esc key in the matrix
+#define MATRIX_ESC_INPUT    7
+#define MATRIX_ESC_OUTPUT   7
+
+// Position of physical Fn key in the matrix
+#define MATRIX_FN_INPUT     3
+#define MATRIX_FN_OUTPUT    17
+
 #endif // _BOARD_KEYMAP_H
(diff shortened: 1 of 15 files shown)


jackpot51 (approved):

jacobgkau (approved):
Key combination works for resetting the battery thresholds.

Some information that may be useful to support as this combination starts getting used:
- The keys must be held down _before_ pushing the power button (pushing the power button and then the keys immediately after does not work.) This should probably be emphasized in any documentation for this feature.
- If Fn-Esc continues being held down after the reset, the UEFI setup menu is displayed (because the key for that is Esc.) Seeing the UEFI menu is not an indication that the reset worked-- there is a brief period after power-on where Fn-Esc is ignored, but after that, it would show the menu without resetting.

If we're not planning to have any sort of feedback like an "EC configuration has been reset, press Enter to continue" message, then this function is doing what it needs to.

jackpot51 (commented):
Just realized that because we will want to reset any sort of keymap, we will probably need to have a different test for Fn+Esc that relies on their physical position instead of their mapping

src/board/system76/addw2/include/board/keymap.h
```diff
@@ -46,4 +46,8 @@
 // Keymap
 extern uint16_t __xdata KEYMAP[KM_LAY][KM_OUT][KM_IN];
 
+// Physical location of the Esc and Fn keys
+#define K_PHYS_ESC KEYMAP[0][7][7]
```
jackpot51, line 50: Two keys could be set to the same value. I'd recommend defining something like ESC_INPUT and ESCAPE_OUTPUT so it tests i and j against them

crawfxrd:
Now checks the matrix position, rather than the key _at_ that position.

jackpot51 (approved):

jacobgkau (approved):

PR #111, opened 2020-09-28
Refactor SMFI and ectool

jackpot51:
ectool now uses /dev/port with POSIX advisory locks for the port range required. All public interfaces are documented


tool/src/access/lpc/linux.rs
```diff
@@ -0,0 +1,183 @@
...
+use super::*;
+
+struct PortLock {
+    start: u16,
+    len: u16,
+    file: fs::File,
+}
+
+impl PortLock {
+    pub fn new(start: u16, end: u16) -> io::Result<Self> {
```
ids1024, line 32: Maybe this should be `start` and `len`, since the caller has a start and a length, and the code here has to convert it to a length anyway.

Not hugely important though.

jackpot51: The reason I chose end is to avoid potential overflows

crawfxrd (approved):

PR #114, opened 2020-10-01
Keyboard configuration saved in flash

jackpot51:
Before this PR, turning off a laptop and unplugging the AC adapter would clear custom mapping.
Afther this PR, mappings should be preserved through such an event when.

Test with https://github.com/pop-os/keyboard-configurator/pull/1

### DIFF
--- a/src/board/system76/common/smfi.c
+++ b/src/board/system76/common/smfi.c
@@ -153,33 +153,35 @@ static enum Result cmd_keymap_get(void) {
     int layer = smfi_cmd[SMFI_CMD_DATA];
     int output = smfi_cmd[SMFI_CMD_DATA + 1];
     int input = smfi_cmd[SMFI_CMD_DATA + 2];
-
-    if (layer < KM_LAY && output < KM_OUT && input < KM_IN) {
-        uint16_t key = KEYMAP[layer][output][input];
+    uint16_t key = 0;
+    if (keymap_get(layer, output, input, &key)) {
         smfi_cmd[SMFI_CMD_DATA + 3] = (uint8_t)key;
         smfi_cmd[SMFI_CMD_DATA + 4] = (uint8_t)(key >> 8);
         return RES_OK;
+    } else {
+        return RES_ERR;
     }
-
-    // Failed if keyboard mapping not found
-    return RES_ERR;
 }
 
 static enum Result cmd_keymap_set(void) {
     int layer = smfi_cmd[SMFI_CMD_DATA];
     int output = smfi_cmd[SMFI_CMD_DATA + 1];
     int input = smfi_cmd[SMFI_CMD_DATA + 2];
-
-    if (layer < KM_LAY && output < KM_OUT && input < KM_IN) {
-        uint16_t key =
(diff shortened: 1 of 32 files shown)


jackpot51:
Before this PR, turning off a laptop and unplugging the AC adapter would clear custom mapping.
Afther this PR, mappings should be preserved through such an event when.

Test with https://github.com/pop-os/keyboard-configurator/pull/1

src/board/system76/common/smfi.c
```diff
@@ -184,33 +184,35 @@ static enum Result cmd_keymap_get(void) {
...
-            ((uint16_t)smfi_cmd[5]) |
-            (((uint16_t)smfi_cmd[6]) << 8);
-        KEYMAP[layer][output][input] = key;
-        return RES_OK;
+    uint16_t key =
+        ((uint16_t)smfi_cmd[5]) |
+        (((uint16_t)smfi_cmd[6]) << 8);
+    //TODO: consider only setting if the key has changed
+    if (keymap_set(layer, output, input, key)) {
+        //TODO: should we write config on every change?
```
ids1024, line 206: Maybe something like:
* Keyboard Configurator sends a command to sync to ROM when it closes cleanly
* Sync periodically (every few minutes?) when there are changes
* Sync, if there are changes, on poweroff/suspend

I guess disk caching in an OS like Linux works something like that, though I don't know the exact approach.

jackpot51: Indeed. It doesn't seem to cause delays though to keep writing to flash, and it will have the same 100,000 change limitation as Launch has. So if there is an apply button just for preventing flash wear, it should probably happen on launch as well.

src/common/include/common/keymap.h
```diff
@@ -3,10 +3,33 @@
...
 #include <stdint.h>
 
+// Keymap defined by board
 #if defined(KM_LAY) && defined(KM_OUT) && defined(KM_IN)
-extern uint16_t __xdata KEYMAP[KM_LAY][KM_OUT][KM_IN];
+    extern uint16_t __code KEYMAP[KM_LAY][KM_OUT][KM_IN];
+    extern uint16_t __xdata DYNAMIC_KEYMAP[KM_LAY][KM_OUT][KM_IN];
+    #define HAVE_KEYMAP 1
+#else
+    #define HAVE_KEYMAP 0
```
crawfxrd, line 15: Should this be an error? Compilation will fail if `KM_*` are not defined.

jackpot51: Good idea. So long as the AVR boards still build. They shouldn't be including keymap.h anyways

jackpot51: Erm, it can't be done this way. common/keymap.c includes some functions that are always present even on AVR boards

jackpot51:
@leviport the only changes done here yesterday were to rebase it on master as other PRs are merged.

crawfxrd (approved):

leviport (approved):
Looking good on darp6 and oryp6

PR #117, opened 2020-10-03
lemp9: Added ISO keyboard layout "default_iso"

fabiant7t:
I flashed the System76 firmware and EC on my Clevo L141CU (great work, thanks a lot!). Unlike the Lemur Pro with its US ANSI layout, mine has a US ISO layout (see https://drive.google.com/file/d/1vTJvdfl9tf2Iwb6hkmsUwGBCVZ9lmouv/view?usp=sharing) and both backslash/pipe keys are malfunctioning after flashing the default ANSI layout. So I created a keyboard matrix mapping `LAYOUT_ISO` and implemented the layout that got printed on the keycaps as `default_iso.c`.

The yet undefined keycodes K_PAUSE, K_SYS_RQ, K_BREAK and K_SCROLL_LOCK can be set once defined, I updated and added the appropriate TODO lines.


fabiant7t:
For reference, here's an annotated and colored photo ot the keyboard and layout: ISO_keyboard_layout_clevo_L141CU

jackpot51:
There are a couple issues I see with this.

Firstly, the System76 Lemur Pro does not use the same mapping that other OEM's of the same chassis are using. We have a custom mapping which is incompatible. So comparing your ISO mapping with our mapping and calling it the default ISO mapping of the Lemur Pro is incorrect. Other OEMs shipping the ISO layout can make mapping modifications just like we did with our mapping. I would recommend naming it differently, to identify the OEM that provided this keyboard mapping. Using LAYOUT and LAYOUT_ISO is still fine, as the layout to electrical pins would still be the same even if keys are remapped.

Secondly, there should be a way to detect that an ISO keyboard is attached. On several of our laptops, we have removable keyboards and have sometimes offered an ISO layout. I intend to support, for example, the UK ISO layout that we shipped on some System76 Darter Pro's soon. Having the ability to dynamically detect the layout can be very beneficial for those systems where the keyboard can be swapped, so the same firmware can support both keyboard layouts.

fabiant7t:
> Firstly, the System76 Lemur Pro does not use the same mapping that other OEM's of the same chassis are using. We have a custom mapping which is incompatible. So comparing your ISO mapping with our mapping and calling it the default ISO mapping of the Lemur Pro is incorrect. Other OEMs shipping the ISO layout can make mapping modifications just like we did with our mapping. I would recommend naming it differently, to identify the OEM that provided this keyboard mapping. Using LAYOUT and LAYOUT_ISO is still fine, as the layout to electrical pins would still be the same even if keys are remapped.

True. Are you fine with renaming it to `clevo_l141cu_us_iso.c`?

> Secondly, there should be a way to detect that an ISO keyboard is attached. On several of our laptops, we have removable keyboards and have sometimes offered an ISO layout. I intend to support, for example, the UK ISO layout that we shipped on some System76 Darter Pro's soon. Having the ability to dynamically detect the layout can be very beneficial for those systems where the keyboard can be swapped, so the same firmware can support both keyboard layouts.

Good idea! Unfortunately, I cannot detect it out of the box. The keyboard device IDs of the Clevo L141CU (ISO) and the Lemur Pro (ANSI) are identical ([Reference](https://linux-hardware.org/index.php?view=computers&vendor=System76&model=Lemur+Pro), [Mine](https://linux-hardware.org/?probe=5f01c32134)). Second best option might be adding the used keyboard layout to the DMI table (as version?). Swapping the keyboard and changing the layout would require to reflash the firmware. But it's probably less common to swap the layout than to swap the keyboard after a spill.

jackpot51:
> > Firstly, the System76 Lemur Pro does not use the same mapping that other OEM's of the same chassis are using. We have a custom mapping which is incompatible. So comparing your ISO mapping with our mapping and calling it the default ISO mapping of the Lemur Pro is incorrect. Other OEMs shipping the ISO layout can make mapping modifications just like we did with our mapping. I would recommend naming it differently, to identify the OEM that provided this keyboard mapping. Using LAYOUT and LAYOUT_ISO is still fine, as the layout to electrical pins would still be the same even if keys are remapped.
> 
> True. Are you fine with renaming it to `clevo_l141cu_us_iso.c`?
I suppose that is fine. Do you not have an OEM that you bought this unit from? Clevo does not sell directly often.
> 
> > Secondly, there should be a way to detect that an ISO keyboard is attached. On several of our laptops, we have removable keyboards and have sometimes offered an ISO layout. I intend to support, for example, the UK ISO layout that we shipped on some System76 Darter Pro's soon. Having the ability to dynamically detect the layout can be very beneficial for those systems where the keyboard can be swapped, so the same firmware can support both keyboard layouts.
> 
> Good idea! Unfortunately, I cannot detect it out of the box. The keyboard device IDs of the Clevo L141CU (ISO) and the Lemur Pro (ANSI) are identical ([Reference](https://linux-hardware.org/index.php?view=computers&vendor=System76&model=Lemur+Pro), [Mine](https://linux-hardware.org/?probe=5f01c32134)). Second best option might be adding the used keyboard layout to the DMI table (as version?). Swapping the keyboard and changing the layout would require to reflash the firmware. But it's probably less common to swap the layout than to swap the keyboard after a spill.
This wouldn't be a DMI thing or something the OS can detect. It doesn't look like the Lemur Pro has any GPIOs to detect the difference.

fabiant7t:
> > > Firstly, the System76 Lemur Pro does not use the same mapping that other OEM's of the same chassis are using. We have a custom mapping which is incompatible. So comparing your ISO mapping with our mapping and calling it the default ISO mapping of the Lemur Pro is incorrect. Other OEMs shipping the ISO layout can make mapping modifications just like we did with our mapping. I would recommend naming it differently, to identify the OEM that provided this keyboard mapping. Using LAYOUT and LAYOUT_ISO is still fine, as the layout to electrical pins would still be the same even if keys are remapped.
> > 
> > 
> > True. Are you fine with renaming it to `clevo_l141cu_us_iso.c`?
> 
> I suppose that is fine. Do you not have an OEM that you bought this unit from? Clevo does not sell directly often.

Great! It's an unbranded Clevo bought at laptopwithlinux.com. No OEM.

> > > Secondly, there should be a way to detect that an ISO keyboard is attached. On several of our laptops, we have removable keyboards and have sometimes offered an ISO layout. I intend to support, for example, the UK ISO layout that we shipped on some System76 Darter Pro's soon. Having the ability to dynamically detect the layout can be very beneficial for those systems where the keyboard can be swapped, so the same firmware can support both keyboard layouts.
> > 
> > 
> > Good idea! Unfortunately, I cannot detect it out of the box. The keyboard device IDs of the Clevo L141CU (ISO) and the Lemur Pro (ANSI) are identical ([Reference](https://linux-hardware.org/index.php?view=computers&vendor=System76&model=Lemur+Pro), [Mine](https://linux-hardware.org/?probe=5f01c32134)). Second best option might be adding the used keyboard layout to the DMI table (as version?). Swapping the keyboard and changing the layout would require to reflash the firmware. But it's probably less common to swap the layout than to swap the keyboard after a spill.
> 
> This wouldn't be a DMI thing or something the OS can detect. It doesn't look like the Lemur Pro has any GPIOs to detect the difference.

jackpot51 (approved):

Issue #103, opened 2020-09-21
Implement max fan keypress


jackpot51:
This used to be fn+1. It could be the start of a manual fan control framework

WatchMkr:
I recall regular support request from customers accidentally hitting the FN+1 combination.

Understanding how it might be used would help decide the best way to provide a max-fan feature (if at all).

jackpot51:
Support has stated that they prefer the small number of complaints about accidentally hitting it to being unable to test fans easily. It is also possible that having support for writing fan duty cycle from the OS would solve this problem, but is harder to implement

WatchMkr:
Got it. Let's use a three-key combo to allow testing and avoid accidentally turning on the fans to max. FN+Shift+1?

jackpot51:
I like two handed keypresses to avoid issues. Any thoughts on a right handed key that could be used?

WatchMkr:
Either side could be used with Shift in the combo. Keeping FN and 1 is good so that if it's not one (old stuff) add Shift (new stuff).

curiousercreative:
Was this issue resolved?

jackpot51:
Yes, it was resolved. Fn-1 triggers maximum fan speed

PR #118, opened 2020-10-05
Implement fan toggle key and refactor duplicate fan code

jackpot51:
Pressing Fn+1 will set the fans to maximum. Reboot, suspend, or shutdown should reset them to automatic control. Partially addresses #103. Requiring shift to be pressed is possible, but makes remapping somewhat nonsensical. If there is a reasonable way to explain this requirement in the keyboard configurator, I can adjust this PR.


src/board/system76/common/fan.c
```diff
@@ -0,0 +1,8 @@
+#include <board/fan.h>
```
crawfxrd, line 1: Missing license ID

crawfxrd (approved):

Issue #122, opened 2020-10-06
galp4: Implement numpad


jackpot51:
Make sure to add a note to the [transition document](https://github.com/system76/docs/blob/54eca9bb610ae7443d9e8f53bddd868bd3d9a984/_articles/transition-firmware.md): "Galago Pro number pad keys always require pressing the function key"

```
Proprietary EC firmware:
            | Function off | Function on
Numlock off | Sends letter | Sends numpad
Numlock on  | Sends numpad | Sends letter

System76 EC firmware:
            | Function off | Function on
Numlock off | Sends letter | Sends numpad
Numlock on  | Sends letter | Sends numpad

Operating System:
Numlock off | Numpad is parsed as navigation key
Numlock on  | Numpad is parsed as number key
```

PR #123, opened 2020-10-07
galp4: Add numpad keys to Fn layer

crawfxrd:
Fn must be held to use the numpad, regardless of the state of NumLock. This differs from the proprietary EC, which swaps the numpad and letters when NumLock is enabled.

The OS determines if the numpad is used as navigation keys or number keys, depending on the state of NumLock.

Resolves: #122


jackpot51 (approved):

jacobgkau (approved):
While it seems awkward to have the numlock keys act as navigation keys when numlock is disabled (which it is by default) since those functions are not printed onto the keyboard, I see in the keymap that we only define them as numlock number keys and we're not explicitly defining them as navigation keys, so this seems like correct behavior and matches the description of what we're doing.

(I looked at enabling numlock by default using the gsetting to make this less awkward, but we can't do that, since it would also enable numlock by default for proprietary EC users, which is much worse than Open EC users having to turn numlock on.)

PR #127, opened 2020-10-21
Add magic strings read by uecflash

crawfxrd:
uecflash, the program used to flash the EC from firmware-update, expects the binary to have specfic embedded strings in order to report the EC project, version, and date.


jackpot51:
Since the protocol was implemented in https://github.com/system76/firmware-update/pull/40, uecflash will never be used to flash System76 EC. And it shouldn't be done due to it resetting to the new firmware without checking for the proper entry point. So, this is now obsolete

PR #129, opened 2020-11-02
Add Custom Valentin French ISO Layout definition

c4software:
Hi,

Since its possible to use the System76 Lemp9 EC on other based L140CU Hardware (in my case the Venus PW55), I have created a custom layout for my French ISO keyboard layout. I have also swap some keys to my personal use.

Thanks again for your work 👌


crawfxrd (approved):

crawfxrd:
We should come up with a scheme for adding keymaps. Probably follow what QMK does.

PR #130, opened 2020-11-05
Sort object files

crawfxrd:
ROMs produced of Pop!_OS 20.10 and Fedora 33 with SDCC 4.0.0 cause uecflash to crash after verifying the block write, leaving the  system unbootable until all power is removed to force an EC reset.

Sorting object files appears to resolve this issue, creates a smaller binary (~200 bytes from 20.04), and gives more reproducible builds.

Resolves: system76/firmware-update#26


crawfxrd:
ROMs built on Fedora 33, Pop!_OS 20.10, and Pop!_OS 20.04 (when using sdcc 4.0.0 from groovy) all have the same SHA256 checksum.

jackpot51 (approved):

PR #131, opened 2020-12-04
ectool: AccessLpcSim for communicating with ecsim

ids1024:
With this and https://github.com/system76/ecsim/pull/1, `system76_ectool info` at least can run, after changing `main.rs` to use `AccessLpcSim` insead of `AccessLpcLinux`.


ids1024:
`keyboard-configurator` also works with ecsim, if patched to use `AccessLpcSim`.

jackpot51 (approved):

PR #136, opened 2021-01-05
LED commands

jackpot51:
Commands will be used by keyboard-configurator in https://github.com/pop-os/keyboard-configurator/pull/22


ids1024:
With this API, a client can determine if there's a controllable keyboard backlight by checking if the max brightness is non-zero, right?

Is there anyway to determine whether or not RGB color is supported, since some models may have a monochrome backlight?

jackpot51:
@ids1024 I could adjust the led_get_color API to return a bool for whether the color can be changed, if that works

jackpot51:
I'm running quite low on EC space on the Launch keyboard so I think it would be best if the keyboard definition said wether it supported color

ids1024:
> so I think it would be best if the keyboard definition said wether it supported color

I suppose that works too, since it's already assumed the configurator includes specifications for different keyboard models.

src/board/system76/oryp5/kbled.c
```diff
@@ -1,64 +0,0 @@
...
-
-void kbled_set(uint8_t level) {
-    kbled_i2c_set(0x12, &level, 1);
-}
-
-void kbled_set_color(uint32_t color) {
-    // Set blue component
-    uint8_t value = (uint8_t)(color);
-    kbled_i2c_set(0x02, &value, 1);
-    kbled_i2c_set(0x05, &value, 1);
```
crawfxrd, line 50: The addresses for oryp5 are different from darp5.

src/board/system76/common/kbled/rgb_pwm.c
```diff
@@ -19,11 +19,28 @@ uint8_t kbled_get(void) {
...
 }
 
+uint32_t kbled_get_color(void) {
+    // Get PWM of blue component
+    uint32_t color = (uint32_t)DCR7;
+
+    // Get PWM of green component
+    color |= ((uint32_t)DCR6) << 8;
+
+    // Get PWM of green component
```
crawfxrd, line 38: red component

src/board/system76/common/kbled/rgb_i2c.c
```diff
@@ -39,10 +39,31 @@ uint8_t kbled_get(void) {
...
 }
 
+uint32_t kbled_get_color(void) {
+    // Get blue component from channel 0
+    uint8_t value;
+    kbled_i2c_get(0x02, &value, 1);
+    uint32_t color = (uint32_t)value;
+
+    // Get green component from channel 3
+    kbled_i2c_get(0x05, &value, 1);
```
crawfxrd, line 57: Same issue as set_color for darp5/oryp5: addresses for color components are different.

src/board/system76/oryp5/kbled.c
jackpot51: Thanks for finding this, I'll try and adjust

jackpot51: I split the two into a darp5 and an oryp5 kbled file. Hopefully Clevo has given up on this strange design so we mostly use rgb_pwm in the future

ids1024:
Another useful feature here, that I had working through https://github.com/pop-os/system76-power/pull/190 (using inotify), is notification of changes. So the Configurator color and brightness could show changes made through keybindings or other methods.

Though that may complicate the API here too much.

jackpot51:
@ids1024 we can't really do that without a kernel driver to handle interrupts. I recommend that the configurator polls at a reasonable interval, maybe 1 second?

PR #139, opened 2021-01-19
fan smoothing and syncing

curiousercreative:
## NOTES
- I'm a C newb and very interested in how this could be refactored for elegance, accuracy, etc.
- First PR on this repo. I developed this off the commit for my current firmware (galp5 2020-12-15_79d6fa7) and later rebased to master. Please advise improvements to the PR itself.

## CHANGES
I wrote this for my new galp5 to address a common complaint of the fans jumping from 0 to 100 and back abruptly. This PR adds support for new build flags:
- `SYNC_FANS` will read target fan speeds from both CPU and GPU fan curves, take the max and set both fans to this speed.
- `FAN_SMOOTHING` sets the number of fan speed event loops that a 0 - 100 speed ramp should be smoothed across. Effectively, this determines the maximum fan speed adjustment for each event loop to prevent audible stepping. A high number yields a slower and smoother ramp while a low number yields a fast response.
- `FAN_SMOOTHING_UP` sets the fan smoothing (as described above) for the fan speed increases
- `FAN_SMOOTHING_DOWN` sets the fan smoothing (as described above) for the fan speed decreases

See updated galp5 board config for an example of these new values: https://github.com/system76/ec/pull/139/files#diff-198f0231dadddcc1692e1557f98782a282a0e540eb1664fd95402cb0574caab8R43

Attempts to fix https://github.com/system76/firmware-open/issues/149 and https://github.com/system76/firmware-open/issues/148

### DIFF
--- a/src/board/system76/common/dgpu.c
+++ b/src/board/system76/common/dgpu.c
@@ -53,7 +53,7 @@ static struct Fan __code FAN = {
     .heatup_size = ARRAY_SIZE(FAN_HEATUP),
     .cooldown = FAN_COOLDOWN,
     .cooldown_size = ARRAY_SIZE(FAN_COOLDOWN),
-    #ifdef FAN_SMOOTHING
+    #if defined(FAN_SMOOTHING) || defined(FAN_SMOOTHING_UP) || defined(FAN_SMOOTHING_DOWN)
       .interpolate = true,
     #else
       .interpolate = false,
@@ -93,7 +93,7 @@ void dgpu_event(void) {
         // Apply heatup and cooldown filters to duty
         duty = fan_heatup(&FAN, duty);
         duty = fan_cooldown(&FAN, duty);
-        #ifdef FAN_SMOOTHING
+        #if defined(FAN_SMOOTHING) || defined(FAN_SMOOTHING_UP) || defined(FAN_SMOOTHING_DOWN)
           duty = fan_smooth(last_duty_dgpu, duty);
           last_duty_dgpu = duty;
         #endif
--- a/src/board/system76/common/dgpu.c
+++ b/src/board/system76/common/dgpu.c
@@ -27,7 +27,6 @@ static uint8_t FAN_HEATUP[BOARD_DGPU_HEATUP] = { 0 };
 static uint8_t FAN_COOLDOWN[BOARD_DGPU_COOLDOWN] = { 0 };
 
 int16_t dgpu_temp = 0;
-uint8_t last_duty_dgpu = 0;
 
 #define DGPU_TEMP(X) ((int16_t)(X))
 
@@ -65,7 +64,7 @@ void dgpu_init(void) {
     i2c_reset(&I2C_DGPU, true);
 }
 
-void dgpu_event(void) {
+uint8_t dgpu_get_fan_duty(void) {
     uint8_t duty;
     if (power_state == POWER_STATE_S0 && gpio_get(&DGPU_PWR_EN) && !gpio_get(&GC6_FB_EN)) {
         // Use I2CS if in S0 state
@@ -93,22 +92,18 @@ void dgpu_event(void) {
         // Apply heatup and cooldown filters to duty
         duty = fan_heatup(&FAN, duty);
         duty = fan_cooldown(&FAN, duty);
-        #if defined(FAN_SMOOTHING) || defined(FAN_SMOOTHING_UP) || defined(FAN_SMOOTHING_DOWN)
-          duty = fan_smooth(last_duty_dgpu, duty);
-          last_duty_dgpu = duty;
-        #endif
     }
 
-    if (duty != DCR4) {
-        DCR4 = duty;
-        DEBUG("DGPU temp=%d = %d\n", dgpu_temp, duty);
-    }
+    DEBUG("DGPU temp=%d\n", dgpu_temp);
+    return duty;
 }
 
 #else
 
 void dgpu_init(void) {}
 
-void dgpu_event(void) {}
+uint8_t dgpu_get_fan_duty(void) {
+  return PWM_DUTY(0);
+}
 
 #endif // HAVE_DGPU
--- a/src/board/system76/common/dgpu.c
+++ b/src/board/system76/common/dgpu.c
@@ -27,6 +27,7 @@ static uint8_t FAN_HEATUP[BOARD_DGPU_HEATUP] = { 0 };
 static uint8_t FAN_COOLDOWN[BOARD_DGPU_COOLDOWN] = { 0 };
 
 int16_t dgpu_temp = 0;
+uint8_t last_duty_dgpu = 0;
 
 #define DGPU_TEMP(X) ((int16_t)(X))
 
@@ -52,7 +53,11 @@ static struct Fan __code FAN = {
     .heatup_size = ARRAY_SIZE(FAN_HEATUP),
     .cooldown = FAN_COOLDOWN,
     .cooldown_size = ARRAY_SIZE(FAN_COOLDOWN),
-    .interpolate = false,
+    #ifdef FAN_SMOOTHING
+      .interpolate = true,
+    #else
+      .interpolate = false,
+    #endif
 };
 
 void dgpu_init(void) {
@@ -88,6 +93,10 @@ void dgpu_event(void) {
         // Apply heatup and cooldown filters to duty
         duty = fan_heatup(&FAN, duty);
         duty = fan_cooldown(&FAN, duty);
+        #ifdef FAN_SMOOTHING
+          duty = fan_smooth(last_duty_dgpu, duty);
+          last_duty_dgpu = duty;
+        #endif
     }
 
     if (duty != DCR4) {
--- a/src/board/system76/common/dgpu.c
+++ b/src/board/system76/common/dgpu.c
@@ -1,10 +1,9 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
 #include <board/dgpu.h>
+#include <board/fan.h>
 
 #if HAVE_DGPU
-
-#include <board/fan.h>
 #include <board/gpio.h>
 #include <board/power.h>
 #include <common/debug.h>
--- a/src/board/system76/common/fan.c
+++ b/src/board/system76/common/fan.c
@@ -25,6 +25,9 @@ uint8_t last_duty_peci = 0;
 
   const uint8_t max_jump_up = (max_speed - min_speed) / (uint8_t) FAN_SMOOTHING_UP;
   const uint8_t max_jump_down = (max_speed - min_speed) / (uint8_t) FAN_SMOOTHING_DOWN;
+#else
+  const uint8_t max_jump_up = max_speed - min_speed;
+  const uint8_t max_jump_down = max_speed - min_speed;
 #endif
 
 void fan_reset(void) {
@@ -122,34 +125,32 @@ uint8_t fan_cooldown(const struct Fan * fan, uint8_t duty) __reentrant {
     return highest;
 }
 
-#ifdef FAN_SMOOTHING
-  uint8_t fan_smooth(uint8_t last_duty, uint8_t duty) __reentrant {
-    uint8_t next_duty = duty;
+uint8_t fan_smooth(uint8_t last_duty, uint8_t duty) __reentrant {
+  uint8_t next_duty = duty;
 
-    // ramping down
-    if (duty < last_duty) {
-      // out of bounds (lower) safeguard
-      uint8_t smoothed = last_duty < min_speed + max_jump_down
-        ? min_speed
-        : last_duty - max_jump_down;
+  // ramping down
+  if (duty < last_duty) {
+    // out of bounds (lower) safeguard
+    uint8_t smoothed = last_duty < min_speed + max_jump_down
+      ? min_speed
+      : last_duty - max_jump_down;
 
-      if (smoothed > duty) {
-        next_duty = smoothed;
-      }
+    if (smoothed > duty) {
+      next_duty = smoothed;
     }
-
-    // ramping up
-    if (duty > last_duty) {
-      // out of bounds (higher) safeguard
-      uint8_t smoothed = last_duty > max_speed - max_jump_up
-        ? max_speed
-        : last_duty + max_jump_up;
-
-      if (smoothed < duty) {
-        next_duty = smoothed;
-      }
-    }
-
-    return next_duty;
   }
-#endif
+
+  // ramping up
+  if (duty > last_duty) {
+    // out of bounds (higher) safeguard
+    uint8_t smoothed = last_duty > max_speed - max_jump_up
+      ? max_speed
+      : last_duty + max_jump_up;
+
+    if (smoothed < duty) {
+      next_duty = smoothed;
+    }
+  }
+
+  return next_duty;
+}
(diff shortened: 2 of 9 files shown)


curiousercreative:
## Sensor graphs
### 2 minutes of prime95 single thread
2 minutes of prime95 single thread

### Geekbench 5 CPU benchmark run
Geekbench 5 CPU benchmark run

## board.mk values
```make
# sync GPU fan speed to CPU fan speed
CFLAGS+=-DSYNC_FANS=1
# smooth the fan speed updates such that 0 to full speed happens over this period (divide by 4 for seconds)
CFLAGS+=-DFAN_SMOOTHING=50

# Custom fan curve
CFLAGS+=-DBOARD_HEATUP=3
CFLAGS+=-DBOARD_COOLDOWN=10
CFLAGS+=-DBOARD_FAN_POINTS="\
	FAN_POINT(65, 25), \
	FAN_POINT(70, 50), \
	FAN_POINT(75, 75), \
	FAN_POINT(80, 100), \
"
```

ZeddieXX:
I think you have a better solution than mine.

If I'm understanding you correctly, even FAN_POINT(80, 100) shouldn't actually hit 100% fan speeds at 80C if it starts ramping up slowly and smoothly, which may bring the temps back down before it gets too hot.

Along with the boot entries/nvram issue and this improved fan curve, I am looking forward to the next firmware update. :)

curiousercreative:
@ZeddieXX on a machine with an adequate cooling system (like a big tower), this might be able to prevent the CPU from ever hitting 80C, but tiny laptops like our Galago Pros lack the cooling capabilities to do the same. In reality, our CPUs exceed 80C within a second! No joke, 40C to 80C in a second. My fan points do their best to keep the CPU below 80C, but under load they'll still dance around the 88C thermal ceiling of "balanced" power mode.

curiousercreative:
For myself, I've found that I prefer a quicker ramp up and a slower ramp down of fan speeds and [this commit](https://github.com/curiousercreative/ec/commit/b39bb7fb3e9a29b244fb786e5326f770de34844c) allows for that configuration. Happy to include that in this PR if it's not too messy.

jackpot51:
I generally like this. Is there a reason to re-enable interpolation? Interpolation was disabled so that fan noises would be more steady under load, with this change I would think it would oscillate upwards and downwards

curiousercreative:
@jackpot51 in my experience, there's nothing more annoying than a constant workload that sits on a fan point, causing the fan speed to infinitely loop between lower speed (thereby increasing temp) and higher speed (decreasing temp). You are of course correct that with interpolation turned on, a non-constant workload runs a high risk of fluctuating but I'd argue that with the default fan points, you're unlikely to have the fans on for an extended duration without the presence of a fairly constant workload.

Ultimately, the interpolation doesn't NEED to be tied to smoothing and could be broken out to its own flag.

curiousercreative:
@jackpot51 revisiting this today as I read other redditors wish they had smoother fans. What do you think about including this commit in the PR: https://github.com/curiousercreative/ec/commit/b39bb7fb3e9a29b244fb786e5326f770de34844c?

jackpot51:
@curiousercreative Yes, you can ass that commit. I'll take a look today. I'd like to replace SYNC_FANS with a test for the GPU being present. I can do that in another commit

jackpot51:
@curiousercreative Yes, you can add that commit. I'll take a look today. I'd like to replace SYNC_FANS with a test for the GPU being present. I can do that in another commit

curiousercreative:
@jackpot51 updated

ZeddieXX:
Will this be merged upstream so all GALP5 owners will get it?  Or is this considered a custom firmware that needs to be compiled and flashed by those who know to seek it out?

curiousercreative:
> Yes, you can add that commit. I'll take a look today. I'd like to replace SYNC_FANS with a test for the GPU being present. I can do that in another commit

@jackpot51 now with a dGPU galp5 in my hands, I think that I'd prefer fan syncing for both iGPU and dGPU galp5 and here's an idea that COULD be satisfactory or ideal for all laptops actually:
1. Sync fans always (I believe Apple does this)
1. Keep separate fan curves for CPU and GPU to allow finer thermal targets
1. Move fan duty setting into the fan module itself while each GPU and CPU/PECI module are responsible for mapping their thermal sensor to a fan duty (as they currently do)
1. fan module will take the max fan duty between CPU and GPU. If GPU is not present, the duty will always be 0, so we'll always be syncing fans to whatever the CPU fan duty is

## Example

### Example GPU fan curve
- 70C = 25%
- 80C = 50%
- 90C = 100%

### Example CPU fan curve
- 65C = 10%
- 70C = 10%
- 72C = 20%
- 80C = 80%
- 88C = 100%

### Mock scenarios
- While GPU = 50C and CPU = 70C, both fans are set to 10%
- While GPU = 70C and CPU = 70C, both fans are set to 25%
- While GPU = 70C and CPU = 60C, both fans are set to 25%
- While GPU = 0C and CPU = 88C, both fans are set to 100%

jackpot51:
I agree, given that all the supported laptops have a shared heatpipe system between the CPU and GPU. Increases in CPU heat should result in spinning the GPU fan a bit more and vice versa.

curiousercreative:
@jackpot51 updated with smarter fan syncing. Feel free to edit for any reason. I've flashed it and so far so good.

curiousercreative:
@jackpot51 anything I can do to get this PR ready for review?

gregor160300:
@curiousercreative this pull request breaks building for any model without a dgpu (maybe even every model other than galp5) 
with:

src/board/system76/common/dgpu.c:106: error 112: function 'PWM_DUTY' implicit declaration
src/board/system76/common/dgpu.c:106: error 101: too many parameters 

I am trying to figure out how to fix it, I'll let you know if I find a fix

curiousercreative:
@gregor160300 thanks! What model was the build failing for (so that I may verify the fix)?

gregor160300:
> @gregor160300 thanks! What model was the build failing for (so that I may verify the fix)?

I tried with both darp7 and lemp10 both failed before only tried again with darp7 after the fix, I did add the  fansmoothing up and down flags to the makefile of darp7 so it might be worth testing also without adding extra cflags

curiousercreative:
@gregor160300 thanks again for testing this out. I've updated with your fix and another that I found.

ZeddieXX:
Does this mean it's ready to be merged? ^_^

jackpot51:
I don't think it uses or needs board_detect, that could be removed.

curiousercreative:
@jackpot51 will do. Do we want to continue defaulting to independent fan speeds and unsmoothed or should we default to synced fans and smoothing? I'd say synced and smoothed makes sense, but is that what S76 wants?

curiousercreative:
rebased without board detect

gregor160300:
I have been using this for a bit now and while it is very nice that there now is a ramp up and ramp down which makes for a much better user experience I don't like that the fan has become quite bouncy. For example when I compile gnome shell git from archlinux aur this means there will be some code compiling then a few seconds of waiting for user input and then more compiling and so on. The problem here is that the fan keeps bouncing between 0 - 5000 rpm (mostly between 4000 - 5000) which isn't great. All the while temps fluctuate between 85 and 55 degrees celcius. As for a proper solution to this the only thing I can currently think of is even more slowly ramping down or even waiting woth ramping down until a certain temperature is reached for at least x seconds.

curiousercreative:
@gregor160300 your experiences are mostly running on AC power? Trouble is how much power consumption can vary for how little thermal mass we have to regulate it. I've observed this CPU jump from 5-50W and 50C to 80C in a second or two. For non-constant workloads (such as what you describe), there's just a moving target that I don't think can be solved for without adding thermal mass. Perhaps the best you can do is what you describe and defaulting to a longer smoothing duration is likely a good option for most use cases. I prefer the window I've got here for performance while I'm docked. I've found it's hard to get my galp5 into fan territory at all on battery power due to a lower power envelope being set.

gregor160300:
@curiousercreative I noticed this behaviour indeed, and for me it is also difficult to get the fans on while on battery (which I deem a good thing) as for making the period without jumps longer I think I can play with the board heatup/cooldown values

curiousercreative:
Try that and increasing fan smoothing.

On February 28, 2021 7:38:54 AM EST, Gregor Schram <notifications@github.com> wrote:
|@curiousercreative I noticed this behaviour indeed, and for me it is
|also difficult to get the fans on while on battery (which I deem a good
|thing) as for making the period without jumps longer I think I can play
|with the board heatup/cooldown values
|
|-- 
|You are receiving this because you were mentioned.
|Reply to this email directly or view it on GitHub:
|https://github.com/system76/ec/pull/139#issuecomment-787445906

Winston Hoy
(202)630-7559
Curiouser Creative Studio
Video, Photo, Web, Motion, Design

barthvader13:
I bought an Oryx Pro last fall and I've written System76 support about this exact issue because the fan curve on my machine is infuriating. Sometimes my fans sound like a rocket ship when I just refresh a web page or open up a random app like Git Kraken. I'll be in random zoom meetings and the fans will spike for absolutely no reason - its literally become a meme with my coworkers now.

System76 Support wrote: "_After updating your firmware, you should have manual control over the fans' turbo speed with the Fn + 1 shortcut, however at this time we still do not have fan curves exposed from within the firmware._"

Is this custom firmware something I could use on my Oryx Pro? lm-sensors and fan-control are not going to work here right?

These are the two links the System76 support person told me to look at:
https://support.system76.com/articles/system-firmware/
https://github.com/system76/firmware-open

I don't want to hear any fans when my CPU is <15%.

jackpot51:
@barthvader13 I referenced your comment in a new issue: https://github.com/system76/ec/issues/163

curiousercreative:
@jackpot51 In digging into some unexpected fan speed fluctuations, I found that fan speed interpolation wasn't actually working properly and tracked it down to the PECI temp bit shifting. Simply reversing that bit shifting seems to have fan speed interpolation working properly again and that's the most recent commit here (along with some addition debug logging that can be reverted).

curiousercreative:
@gregor160300 can you try flashing the latest of this branch? Curious if you'll see less unexpected fan speed fluctuations.

gregor160300:
@curiousercreative I'm sorry, I had to return mine due to unrelated issues.

curiousercreative:
@jackpot51 some fun observations regarding the galp5 fans (from psensor):
- No fan point below 25% spins the fans at all (I tested 10, 15 and 20%)
- 25% fan point (64 of 255) spins CPU fan at about 1000RPM and dGPU at 2000RPM
- 50% fan point (128 of 255) spins CPU fan at about 3180RPM and dGPU at 3830TTtRPM
- 100% fan point (255 of 255) spins CPU fan at about 6178RPM and dGPU fan at about 6304RPM

ZeddieXX:
Any updates on this?  Very excited to see this on the next official EC firmware update.

ZeddieXX:
@jackpot51 I just saw a new firmware just dropped for the GALP5.  It doesn't list smoother fan curves or syncing, however.  Is this implemented?  If not, what is preventing this from going into the production firmware?

jacobgkau:
> Is this implemented?

Released firmware is typically based on the `master` branch at the time of creation. You can tell that this PR has not been merged into master yet because it is still open and in @curiousercreative's `fans-rebase` branch.

jacobgkau:
> Is this implemented?

@ZeddieXX Released firmware is typically based on the `master` branch at the time of creation. You can tell that this PR has not been merged into master yet because it is still open and in @curiousercreative's `fans-rebase` branch.

ZeddieXX:
> 
> 
> > Is this implemented?
> 
> @ZeddieXX Released firmware is typically based on the `master` branch at the time of creation. You can tell that this PR has not been merged into master yet because it is still open and in @curiousercreative's `fans-rebase` branch.

So there's no plan in fan smoothing and syncing for the master branch?

jackpot51:
@curiousercreative I'd like to see the change for syncing CPU and DGPU fan speeds in a separate PR, to make it easier to release these changes

curiousercreative:
@jackpot51 sure, I can work on that. I'll update this PR to include just smoothing and I'll open a new PR based off this one (so including fan smoothing)

jackpot51:
@curiousercreative I would prefer the other way around. A PR that just implements syncing, nothing else. The smoothing is actually far more complicated for me to approve

curiousercreative:
@jackpot51 sure, I can do it that way as well. I understand it being complicated, but I'd bet the fan smoothing is by far the more important of the two features given what I see on the subreddit.

curiousercreative:
@jackpot51 please see https://github.com/system76/ec/pull/177 for fan syncing

curiousercreative:
@jackpot51 speaking of fan smoothing. from this review that was just shared on reddit: https://www.reddit.com/r/linuxhardware/comments/ms8h4k/i_ditched_hp_in_favor_of_system76_here_are_my/
> The fan is usually silent but when it kicks on, it can get quite loud. I wish that when temps increase, the fan slowly trickles up so the noise doesn't feel as abrupt. Display is great, but I didn't test it for color accuracy or anything.

MilesBHuff:
Warning:  This PR broke the CPU and GPU temperature monitors when I merged this to master on my local.  Reverting the merge fixed this issue.
Also, interpolation doesn't seem any less broken with this.

curiousercreative:
@MilesBHuff what's your board.mk look like when you observed the issues? This is what my galp5 is running with good results (including psensor CPU and GPU sensor readings): https://github.com/curiousercreative/ec/blob/galp5/src/board/system76/galp5/board.mk#L44

MilesBHuff:
@curiousercreative https://github.com/MilesBHuff/ec/blob/e0d29fb76e715e7bf70077da981dd6db6e53ec55/src/board/system76/oryp7/board.mk

curiousercreative:
@MilesBHuff did you have the fan smoothing build flags in there when things weren't working?

MilesBHuff:
@curiousercreative Ah Christ, I'm an idiot.  No.  But that shouldn't have affected the CPU/GPU temperature sensors.

curiousercreative:
@MilesBHuff shouldn't, but could be a regression. I don't know if I've ever run the build without the fan smoothing flags, so I can give that a shot at some point. But to your previous comment about interpolation not working, it's only enabled with fan smoothing flags set. You can try it again with smoothing flags and report back.

MilesBHuff:
@curiousercreative I'm mildly afraid to run it knowing it may break the sensors, like before.  The CPU runs very hot, and I'm afraid that the lack of sensor data might impact the ability of the fan curve to function properly.  As well, I've already hard-coded a smooth fan curve in my `board.mk`, so I don't need interpolation anymore.

curiousercreative:
@MilesBHuff ok, no worries. FWIW, I wouldn't worry about running too hot. The CPU package internally has thermal protections, it should be pretty difficult to do it any serious harm, especially in a short time window.

nbs:
I've seen no mention of the Serval WS (serw12 I think) which has atrocious fan noise for no apparent reason. Can this please be ported to that series?

leviport:
>I've seen no mention of the Serval WS (serw12 I think) which has atrocious fan noise for no apparent reason. Can this please be ported to that series?

Sorry, but the Serval WS does not run open EC, so that won't be possible.

curiousercreative:
closing in favor of newer implementation https://github.com/system76/ec/pull/190

PR #141, opened 2021-01-20
Update board.mk

ZeddieXX:
Finer steps for CPU fan for smoother transition.

I don't have a DGPU, but I'm not sure how to make the DGPU fan work like the CPU fan instead.  With both fans working for CPU-only GALP5, I think we can keep the fan noise lower while keeping CPU temps in check.  I would suggest making the a CPU-only version of this firmware to utilize the beefy cooling system to keep the CPU cooler while keeping fan noise down.

For DGPU users, they will need a keep the DGPU fan and CPU fan separated, but I suggest a more granular fan curve (use my CPU fan curve an an example for the DGPU fan for those users).


MilesBHuff:
Should be unnecessary once #139 is merged, since it adds interpolation.  Also, your changes to HEATUP and COOLDOWN completely defeat the point of #43.

PR #146, opened 2021-02-16
Firmware locking

jackpot51:
Part of https://github.com/system76/firmware-open/pull/159

### DIFF
--- a/doc/security.md
+++ b/doc/security.md
@@ -1,0 +1,13 @@
+# Firmware security
+
+The firmware security feature can be configured by setting `CONFIG_SECURITY=1`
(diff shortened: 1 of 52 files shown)


crawfxrd:
Should add a doc describing what `SECURITY` is and how the mechanism works. Would it be better to add to firmware-open instead since it is for locking BIOS regions? (The config should be documented here regardless.)

On that note, should use something more specific that `SECURITY` since the mechanism locks system firmware regions and is not the security of the EC firmware. Does AMD have a mechanism like this or is Intel specific?

crawfxrd (approved):

PR #147, opened 2021-02-18
docs: Explicitly mention to use a USB cable

crawfxrd:
Replace the vague step about grounding with one that says to connect a USB cable and explain its purpose.

### DIFF
--- a/doc/flashing.md
+++ b/doc/flashing.md
@@ -15,18 +15,25 @@ make BOARD=<vendor>/<model> flash_internal
 
 Use this method for first-time flashing or flashing a bricked controller.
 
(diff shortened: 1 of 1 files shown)


jackpot51:
Thing is, that isn't the only way. I frequently use a USB-A cable.

crawfxrd:
If USB-A works I can just say to use a USB cable. I remember you telling me to specifically use a USB-C cable at some point.

jackpot51:
I think that is because USB-C handles host to host connections more better

jackpot51:
How about:

Ground the laptop to the programmer. The recommended way to do this is with a USB-C cable connecting the laptop to the computer doing the proramming

jackpot51 (approved):

PR #156, opened 2021-02-24
Use BIT macro everywhere

crawfxrd:


scripts/coccinelle/bit-macro.cocci
```diff
@@ -0,0 +1,14 @@
...
+@@ expression val; @@
+-(1 << val)
++BIT(val)
+
+@@ expression val; @@
+-(1U << val)
++BIT(val)
+
+@@ expression val; @@
+-1 << val
```
jackpot51, line 13: This one seems potentially dangerous, it could over-match

crawfxrd: I'm not entirely sure how it matches, but it seems to match `1` exactly, as `0b11 << N` values were not modified.

jackpot51 (dismissed):
Just need an approval from @system76/quality-assurance for this. It looks good to me

crawfxrd:
Rebased on master

jackpot51 (approved):

PR #158, opened 2021-02-25
tool: API and CLI for LED matrix mode

ids1024:
See https://github.com/system76/qmk_firmware/pull/6.

### DIFF
--- a/tool/src/ec.rs
+++ b/tool/src/ec.rs
@@ -31,6 +31,8 @@ enum Cmd {
     LedSetValue = 12,
     LedGetColor = 13,
     LedSetColor = 14,
+    LedGetMode = 15,
+    LedSetMode = 16,
 }
 
 const CMD_SPI_FLAG_READ: u8 = 1 << 0;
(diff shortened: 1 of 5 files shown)


jackpot51:
We also need LED `speed`, another uint8_t

ids1024:
Ah, I forgot the speed. Easy enough to add that to the command.

jackpot51:
It should be on the same command, since they heavily rely on eachother

ids1024:
`rgb_matrix_get_speed()`, `rgb_matrix_set_speed()`, and `rgb_matrix_set_speed_noeeprom()` seem to have been added to QMK since our fork was branched off.

tool/src/ec.rs
```diff
@@ -31,6 +31,8 @@ enum Cmd {
     LedSetValue = 12,
     LedGetColor = 13,
     LedSetColor = 14,
+    LedGetMode = 15,
```
jackpot51, line 34: It is good to reserve these in the src/common/include/common/command.h file as well

jackpot51 (approved):

jackpot51 (approved):

PR #159, opened 2021-02-26
Always notify OS of SCI when pmc_sci is called

jackpot51:
This might fix issues with the oryp7 missing SCI events. It will need testing on multiple models and inspection of the kernel logs for errors. I threw in a change to not repeat K_FAN_TOGGLE, which I found while testing this.

### DIFF
--- a/src/board/system76/common/kbscan.c
+++ b/src/board/system76/common/kbscan.c
@@ -275,6 +275,7 @@ static inline bool key_should_repeat(uint16_t key) {
     case K_AIRPLANE_MODE:
     case K_CAMERA_TOGGLE:
     case K_DISPLAY_TOGGLE:
+    case K_FAN_TOGGLE:
     case K_SUSPEND:
     case K_KBD_BKL:
     case K_KBD_COLOR:
--- a/src/board/system76/common/pmc.c
+++ b/src/board/system76/common/pmc.c
@@ -53,16 +53,15 @@ static void pmc_sci_interrupt(void) {
 }
 
 bool pmc_sci(struct Pmc * pmc, uint8_t sci) {
-    // Set SCI queue if possible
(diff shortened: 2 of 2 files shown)


jackpot51:
## Previously

- fwupd
  - [system76-launch: Prompt for unlock keypress if reset command is blocked](https://github.com/fwupd/fwupd/commit/9beb12ad97c6a92ec38261dbc22a129019e6825c)
- keyboard-configurator
  - [Add leds.json file and display index in LEDs tab](https://github.com/pop-os/keyboard-configurator/pull/28)
- qmk_firmware
  - [Unlock on RESET keypress](https://github.com/system76/qmk_firmware/commit/e4adf94e7651cfc33070055a5e0d0bb847821a50)

## Today

- Fixes for oryp7
- Work on launch firmware

crawfxrd (approved):

jacobgkau (approved):
Fan toggle no longer repeats when held down. The SCI queue can't be tested very thoroughly since we couldn't reproduce the issue with it, but it doesn't seem to cause any problems. This will be more fully tested with the next firmware images (and I'll keep an eye out for the issue occurring again.)

After flashing this with `flash.sh`, the splash screen showed up and went away but the OS wouldn't boot and I couldn't access the boot menu until I reset the EC (required unplugging the machine in my case.)

> It will need testing on multiple models and inspection of the kernel logs for errors.

Here is a dmesg log (still with that ACPI AML debug option enabled): 
[dmesg-oryp7.txt](https://github.com/system76/ec/files/6052875/dmesg-oryp7.txt) Not sure if there's anything specific I should look for, but I don't see any problems standing out.

jackpot51:
@jacobgkau let's see if it happens in the USB firmware update

PR #162, opened 2021-03-04
tool: Fix building without `std` feature

ids1024:


ids1024:
I'll add a GitHub action to test this.

crawfxrd:
clap also needs to be made optional so the lib doesn't bring it in.

crawfxrd (approved):

jackpot51 (approved):

jackpot51:
0.3.4 will include this

Issue #163, opened 2021-03-04
Oryx Pro Fan Curve Improvement


jackpot51:
I bought an Oryx Pro last fall and I've written System76 support about this exact issue because the fan curve on my machine is infuriating. Sometimes my fans sound like a rocket ship when I just refresh a web page or open up a random app like Git Kraken. I'll be in random zoom meetings and the fans will spike for absolutely no reason - its literally become a meme with my coworkers now.

System76 Support wrote: "_After updating your firmware, you should have manual control over the fans' turbo speed with the Fn + 1 shortcut, however at this time we still do not have fan curves exposed from within the firmware._"

Is this custom firmware something I could use on my Oryx Pro? lm-sensors and fan-control are not going to work here right?

These are the two links the System76 support person told me to look at:
https://support.system76.com/articles/system-firmware/
https://github.com/system76/firmware-open

I don't want to hear any fans when my CPU is <15%.

_Originally posted by @barthvader13 in https://github.com/system76/ec/issues/139#issuecomment-790885114_

jackpot51:
@barthvader13 please provide the output of `sensors` when the fans are running but the CPU is <15%.

jackpot51:
Also, which graphics mode do you typically use? Do you use external displays?

barthvader13:
First of all, thank you for your help.

I use NVIDIA Graphics mode with the High Performance profile.
I use two external displays: **HDMI** **@1920x1080**  and **USB-C** **@3440x1440**

The reason I use the NVIDIA Graphics mode is because I get screen tearing on the external monitors while using any of the other modes. I fix screen tearing with a startup script that runs
`nvidia-settings --assign CurrentMetaMode="nvidia-auto-select +0+0 { ForceFullCompositionPipeline = On }"`
which only works in NVIDIA Graphics mode.

The biggest culprit for fan noise is obviously Zoom and any video conferencing software like Slack Video or Skype. I keep my mic muted as much as possible in almost every meeting I'm in even 1on1s because of the fan noise. I will get a sensors output for you next time I'm in a zoom meeting. 

My main concern is I wish I had more control over the fan curve because 30% CPU should not cause the fans to be loud enough that my coworkers can hear it. Though I think this issue may be amplified due to the design and placement of the mic on the Oryx Pro.

The other issue I noticed in general is that the fans turn on randomly for about 20-30 seconds at a time with seemingly no correlation to CPU usage/temps - this is the <15% CPU I initially posted about. 

[...]

PR #164, opened 2021-03-05
oryp6, oryp7: Improved fan curve for video chat

jackpot51:
This should help with #163. Fans spin at very low speed at lower temperature to prevent the loud ramp-up that previously happened, especially when on NVIDIA graphics mode and doing something moderately loading the graphics card like video chat.

The system can still be silent under light use, and can still adapt to heavy load (with more noise). I think skin temperature is somewhat better compared to previously, due to running the fan at lower temperatures.

Here is a USB image to use for testing (on the oryp7 only!), unzip it and use Popsicle to flash the .img file to a USB stick, and then boot from the USB stick to flash the firmware:
[oryp7-fan-curve.zip](https://github.com/system76/ec/files/6091893/oryp7-fan-curve.zip)

To test on the oryp6, please build and flash this from source.


jacobgkau (approved):
This seems like a good change, testing on the 17" model with 3060 graphics. The fans are nearly silent at the new 50-59 degree speed. When stressing the system, the fans still spin up after a minute or so (they might take a little longer to spin up than before), but they are quieter under sustained load than before.

When using the system moderately with an external display connected (to try and replicate the problem scenario from before), the system tends to stay under 60 degrees. When it does tip over 60, the fans are still significantly quieter than they were before (more of a whirring than a blasting.)

Would still be good to hear if this addresses what @maria-komarova was seeing during video calls.

maria-komarova:
So far that really helped with the noise. I haven't yet had video calls to test it but I can barely hear fans when not stressing the system.

PR #165, opened 2021-03-09
Reindent files using spaces

crawfxrd:


jackpot51:
I often edit files on the laptop I am building firmware for. Often this means no editor config, and some random editor throws in the wrong whitespace. Perhaps there should be a lint for this?

jackpot51 (approved):

jackpot51 (approved):

PR #167, opened 2021-03-16
oryp6, oryp7: Fix buzzing at lowest fan speed

jackpot51:
This merges the 50C and 60C fan points into a single fan point at 55C with 25% fan duty. By dropping the 20% fan duty point, it should fix fan buzzing described in #163


jacobgkau (approved):
Testing on 17" oryp7, I'm not sure how much of an improvement this is actually making. When testing in a quiet room, I hear what sounds like normal fan noise both before and after the change; that includes a lower buzzing/rumbling in addition to the sound of the air moving.

I don't have an easy way to record externally, but I tried taking some recordings using the internal microphone while stressing one or two CPU cores to generate a very low load. If anything, the fan just seems to get slightly louder slightly quicker after the change. Amplifying and comparing waveforms in Audacity seems to support this observation:

Screenshot from 2021-03-16 13-50-14

I had one recording where one channel had significantly more noise than the other; for the microphone/calling issue, I'm wondering if there's not something else going on with either the microphone or the air blowing over it. (I tried adjusting the lid angle, but didn't find any particular angle that caused more noise.)

In terms of RPM, before the change, I saw the fan start at ~1100, then increase to ~1500. After the change, I'm seeing it start at ~1500 (then sometimes kick up higher to around ~1900.) The faster the fan goes, the quieter (or at least less noticeable) the "buzzing" is.

I'm approving since this seems to do what it's trying to do, but it would be great if we can get one of the users who's particularly affected by this buzzing to confirm if this new curve is acceptable before merging/releasing. (Otherwise, we could end up needing a _third_ fan curve update in a short period of time.)

leviport (approved):
The 15" oryp6 still sounds very quiet. I couldn't hear any buzzing before this change, and I don't hear any after.

jackpot51:
I'm removing approvals pending any of the users who had the buzzing confirming this fixed things

radiantone:
This really improved the sound of my oryx pro! This is a great laptop but the constant max rpm fan when nothing is happening was making this laptop almost unusable.

But not its quiet most of the time and temps are good. Thank you so much.

PR #172, opened 2021-04-02
Fixes for S0iX systems

### LINKED ISSUE #169: galp5 & lemp10: pressure on lid can press keys and wake from suspend
It looks like the keyboard isn't disabled while the lid is closed. I believe the same thing happened on a couple other machines, so hopefully it's a quick fix.

### LINKED ISSUE #170: galp5: fans remain at 100% during suspend
If I set the fans at 100% with the Fn + 1 hotkey and suspend the machine, it will suspend, but the fans stay spinning at 100% while it sleeps. I can also turn them on and off with the machine suspended without it waking up. I think the hotkey should probably be ignored while suspended, and the fans should probably turn off when entering suspend.

jackpot51:
- [x] Fix fan_max keeping fan on when in S0iX - fixes #170 
- [x] Report all keys as released when lid is closed - fixes #169

### DIFF
--- a/src/board/system76/common/dgpu.c
+++ b/src/board/system76/common/dgpu.c
@@ -6,6 +6,7 @@
 
 #include <board/fan.h>
 #include <board/gpio.h>
+#include <board/peci.h>
 #include <board/power.h>
 #include <common/debug.h>
 #include <common/macro.h>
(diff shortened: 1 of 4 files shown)


jacobgkau (approved):
Confirmed both issues are fixed on lemp10 & galp5. Full regression testing will happen with the firmware image.

crawfxrd (approved):

PR #173, opened 2021-04-12
tool: Make `Access` require `Send` and `'static` 

ids1024:
This allows the Configurator to send a `Ec<Box<dyn Access>>` through a channel to a background thread. This could be done differently, but presumably there's no reason to have an `Access` implementation this oesn't apply to.

### DIFF
--- a/scripts/deps.sh
+++ b/scripts/deps.sh
@@ -13,6 +13,7 @@ source /etc/os-release
 
 msg "Installing system build dependencies"
 if [[ "${ID}" =~ "debian" ]] || [[ "${ID_LIKE}" =~ "debian" ]]; then
+    sudo apt-get update
     sudo apt-get install \
         --no-install-recommends \
         --yes \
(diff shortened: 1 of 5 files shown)


ids1024:
I love when CI fails at package installation. I guess if it called `apt update` first, it wouldn't get a 404 downloading packages?

jackpot51 (approved):

jackpot51:
Can you check the failed actions?

.github/workflows/ci.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Install dependencies
-        run: ./scripts/deps.sh
+        run: sudo apt-get update && ./scripts/deps.sh
```
crawfxrd, line 27: Should `deps.sh` just do this itself?

jackpot51: Yes, I think so

ids1024: Don't see any reason not to; updated.

Apparently `dnf install` already updates it's cache, so nothing needs to be changed for it.

crawfxrd (approved):

jackpot51 (approved):

jackpot51:
@ids1024  I just published system76_ectool 0.3.6

PR #177, opened 2021-04-16
Sync dGPU and PECI fans (at highest requested duty)

curiousercreative:
Fix for https://github.com/system76/firmware-open/issues/148

## CHANGES
This PR adds support for new build flags:
- `SYNC_FANS` will read target fan speeds from both CPU and GPU fan curves, take the max and set both fans to this speed.

### DIFF
--- a/src/board/system76/common/dgpu.c
+++ b/src/board/system76/common/dgpu.c
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
 #include <board/dgpu.h>
+#include <board/fan.h>
 
 #if HAVE_DGPU
 
@@ -61,7 +62,7 @@ void dgpu_init(void) {
     i2c_reset(&I2C_DGPU, true);
 }
 
-void dgpu_event(void) {
+uint8_t dgpu_get_fan_duty(void) {
     uint8_t duty;
     if (power_state == POWER_STATE_S0 && gpio_get(&DGPU_PWR_EN) && !gpio_get(&GC6_FB_EN)) {
         // Use I2CS if in S0 state
@@ -91,16 +92,16 @@ void dgpu_event(void) {
         duty = fan_cooldown(&FAN, duty);
     }
 
-    if (duty != DCR4) {
-        DCR4 = duty;
-        DEBUG("DGPU temp=%d = %d\n", dgpu_temp, duty);
-    }
+    DEBUG("DGPU temp=%d\n", dgpu_temp);
+    return duty;
 }
 
 #else
 
 void dgpu_init(void) {}
 
-void dgpu_event(void) {}
+uint8_t dgpu_get_fan_duty(void) {
+  return PWM_DUTY(0);
+}
 
 #endif // HAVE_DGPU
--- a/src/board/system76/common/fan.c
+++ b/src/board/system76/common/fan.c
@@ -9,6 +9,10 @@ bool fan_max = false;
 #define max_speed PWM_DUTY(100)
 #define min_speed PWM_DUTY(0)
 
+#ifndef SYNC_FANS
+  #define SYNC_FANS 1
+#endif
+
 void fan_reset(void) {
     // Do not manually set fans to maximum speed
     fan_max = false;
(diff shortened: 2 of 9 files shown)


jackpot51 (approved):

jackpot51:
This looks good to me, if QA can test on both the Intel and NVIDIA graphics galp5 and approve then I will merge. This will make it easier to read the changes in #139

MilesBHuff:
This works perfectly on the oryp7, which has a dGPU.

jacobgkau:
This seems to be working as intended on galp5 w/ 1650. The dGPU fan spins up whenever the CPU fan spins up, and vice versa. My only concern would be that the speeds don't quite match (the dGPU fan is always slightly faster.)

Samples from stressing the CPU in Hybrid mode:
```
CPU fan:     3769 RPM
GPU fan:     4568 RPM

CPU fan:     3782 RPM
GPU fan:     4568 RPM

CPU fan:     3776 RPM
GPU fan:     4558 RPM

CPU fan:     3769 RPM
GPU fan:     4568 RPM

CPU fan:      426 RPM
GPU fan:      468 RPM
```

Samples from attempting to stress the GPU in Compute mode (the CPU still ended up being warmer than the GPU):
```
CPU fan:     4791 RPM
GPU fan:     5404 RPM

CPU fan:     4791 RPM
GPU fan:     5390 RPM

CPU fan:     4791 RPM
GPU fan:     5404 RPM

CPU fan:     4791 RPM
GPU fan:     5404 RPM

CPU fan:     4791 RPM
GPU fan:     5390 RPM
```

This is strange because the fan part number seems to be the same according to the service manual (6-31-NV40S-104), so while a few RPM of difference might be expected, I would have thought they'd be closer together than this. Are these variances expected?

Aside from that, this should be fine if it's also behaving well on the Intel-only unit that @leviport has.

curiousercreative:
@jacobgkau thanks for QA'ing! I've [observed similar behavior](https://github.com/system76/ec/pull/139#issuecomment-796385323). You can observe it even before this build by setting max fans (fn+1). The dGPU fan always seems to spin higher! Also, the fans don't spin up at all below a certain requested duty....

leviport (dismissed):
I'm seeing the "GPU fan" run just a smidge faster on the Intel-only galp5. Other than that little quirk, everything is looking great!

ZeddieXX:
Yay!  Can't wait to see this in the next EC firmware update!

curiousercreative:
Next question is: should we default to sync for all devices? Should we even have it as a build flag and rather hard code?

MilesBHuff:
> 
> 
> Next question is: should we default to sync for all devices? Should we even have it as a build flag and rather hard code?

I think that depends on whether all devices have a unified CPU/GPU heatpipe.

curiousercreative:
@MilesBHuff I don't know for certain, but @jackpot51 seemed to indicate that to be true [here](https://github.com/system76/ec/pull/139#issuecomment-776907215)

MilesBHuff:
@curiousercreative Ah, interesting!  Cool.  Still, I'd think maybe they'd want to keep their options open for future laptops that don't have a unified heatpipe?  Then again, I can't think of why you wouldn't want one.  

Maybe as a compromise, make this new setting the default, but leave a build flag in case something needs it in the future?

curiousercreative:
@MilesBHuff yeah, that's my inclination as well, but I'd love for S76 to chime in.

curiousercreative:
@jackpot51 any thoughts on whether we want to default fan syncing to on vs off? Anything else that keeps this from being merged? I'm eager to update the fan smoothing pull request

jackpot51:
@curiousercreative yes, let's efault to fan syncing

jackpot51 (approved):
Looks great! Now I'd want @system76/quality-assurance to see if this syncs on the oryp6 or oryp7. If so, I'm good to merge it in and release

MilesBHuff:
> 
> 
> Looks great! Now I'd want @system76/quality-assurance to see if this syncs on the oryp6 or oryp7. If so, I'm good to merge it in and release

It syncs on my oryp7, fwiw.

crawfxrd (approved):

lucserre:
Pardon the n00b question... I can see that this was merged into the master branch now. Amazing!  How does this work now, can we expect that eventually this will be 'pushed' as a firmware update to my oryp7 or is it necessary to manually update?

jacobgkau:
@lucserre Now that it's been merged, this would be included in the next firmware update that you receive.

ZeddieXX:
So glad to see the fan sync merged!  Are we still waiting on the fan smoothing?

MilesBHuff:
@ZeddieXX Yes -- see #190.

curiousercreative:
@jacobgkau @leviport related to our observations of GPU fans running faster, I'm now observing that either pressure to the bottom cover and/or blocking the vents on the bottom cover seems to affect the ACPI reported fan speeds. With fans full blast (fn+1) and machine on your lap (with legs kicked up on a coffee table to be precise), try applying pressure with your left hand (in typing position) and observe one of the two fan speeds jump 2-300 RPM and then try the other hand and observe the a similar change in the other fan.

PR #179, opened 2021-04-20
Fix fan issues with oryp7

MilesBHuff:
The oryp7 (and oryp6, which has the same everything (apart from the GPU)) has abysmal CPU thermals, reaching over 90'C at just 15%.  Yet, in spite of this, and in spite of the fact that the heat pipe is shared with the GPU, only one fan turns on to cool the CPU.  This PR adds the new config option from #177 to the oryp7, thus helping resolve the above issue.

The biggest problem with the fans, though, is undoubtedly the fact that they are not set to turn on until long after CPU throttling has already begun.  Fans should reach full speed _before_ the CPU starts throttling;  and throttling is supposed to be a last resort, not a first one.  In this vein, I have added the vastly improved fan curve from #135.  This will prevent throttling, increase performance, lower temps, and prevent the sudden and harsh spinup so-often complained-about in reviews of the oryp7.


MilesBHuff:
You don't need to merge this PR as-is.  This is largely a call to please implement sane defaults for the fan curves.
This PR works perfectly fine on my oryp7.

curiousercreative:
@MilesBHuff I'd suggest making this a draft PR if it's not ready for review. You can find a "converting to draft" text link in the right sidebar.

MilesBHuff:
Dropped e35f6a5 and squashed f077127.  I'd say this is ready for review now.

MilesBHuff:
I've set the floor speed to 25% to avoid the rattling that happened after #163 was fixed the first time.  Unfortunately, the fan still rattles momentarily when turning off.

I've also halved the cooldown time, to avoid the fan blowing at 100% for over 10 seconds when the CPU is at 45'C (This happens every boot.).  The new cooldown setting works very nicely.

MilesBHuff:
I've just given a full breakdown of the though-processes behind this PR in #180.

MilesBHuff:
Rebased onto latest code (now that #177 has been merge), and added support for smoothing (from #190 -- make sure to merge that before merging this).

MilesBHuff:
Rebased onto latest code (now that #180 has been merged).

crawfxrd:
Fan control has been rewritten, so fan behavior is now completely different.
Now working on adding a manual mode so SBIOS/OS can control target duty.

If you would still like to modify the fan points, I would say it should be limited to 10 points.

PR #184, opened 2021-04-25
Update keyboard customization instructions

larsblumberg:
I overlooked the first paragraph in the keyboard customization readme and jumped directly to the build steps so that I missed to install the build dependencies. Additionally, as `rustup` wasn't installed on my machine, I had to `source $HOME/.cargo/env` because I didn't restart the shell between `./scripts/dev.sh` and `make flash_internal`.

This PR updates the installation instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,3 +18,9 @@ Install dependencies using the provided script:
 ```
 ./scripts/deps.sh
 ```
+
+The dependencies include `rustup` - if it wasn't installed on your system before and you plan to flash the firmware within the same shell session, you should additionally source the environment variables for cargo to avoid build failures:
+
+```
+source $HOME/.cargo/env
+```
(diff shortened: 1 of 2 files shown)


leviport:
I'd rather not have information duplicated in multiple places. That's just a recipe for things not getting updated. If you think the dependencies section is too easy to miss, I guess a note along the lines of "Make sure all dependencies are installed (see above)" would be okay. 

If we're going to include the cargo env sourcing command, I'd prefer that only went into README.md.

larsblumberg:
@leviport I have made the changes as requested by you. Additionally, I've added a paragraph about the sourcing of cargo's env vars to avoid people like me new to rustup/cargo run into build failures.

leviport (approved):
Looks great, thank you for the contribution!

crawfxrd (changes requested):
Please squash the commits as fixups.

crawfxrd (approved):

larsblumberg:
@crawfxrd Squashed!

jackpot51 (approved):

PR #186, opened 2021-04-25
Add missing dependency for `make flash_internal`

larsblumberg:
I had to manually install `cargo` to have `make flash_internal` succeed. This PR adds this dependency to the dependency installation script.

### DIFF
--- a/scripts/deps.sh
+++ b/scripts/deps.sh
@@ -19,6 +19,7 @@ if [[ "${ID}" =~ "debian" ]] || [[ "${ID_LIKE}" =~ "debian" ]]; then
         --yes \
         avr-libc \
         avrdude \
+        cargo \
         curl \
         gcc-avr \
         libudev-dev \
@@ -30,6 +31,7 @@ elif [[ "${ID}" =~ "fedora" ]] || [[ "${ID_LIKE}" =~ "fedora" ]]; then
         avr-gcc \
         avr-libc \
         avrdude \
+        cargo \
         curl \
         make \
         sdcc \


crawfxrd:
The distro-packaged version should not be used. `cargo` gets installed with `rustup`.

larsblumberg:
@crawfxrd I now see why `cargo` wasn't found after running `./scripts/deps.sh`: I run `make flash_internal` in the same bash session and the modified `$PATH` which got updated when installing `rustup` wasn't yet available in the bash session. I could have restarted the shall or sourced `~./.bashrc`, but that's not generic enough for every Linux distribution you want to support.

I'd like to modify this PR to make the updated `$PATH` available in the same shell session where `./scripts/deps.sh` was run. What would be a good way?

larsblumberg:
@crawfxrd When running `./scripts/deps.sh` and `make flash_internal` in the same shell session, `cargo` won't be available yet because `$PATH` isn't yet updated to include the path to `cargo`. That's only the case if the shell session is restarted in between these two commands.

```
$ /scripts/deps.sh
Installing system build dependencies
[...]
info: installing component 'cargo'
[...]

$ cargo
Command 'cargo' not found, but can be installed with:
sudo apt install cargo

$ make flash_internal
[...]
cargo build --manifest-path tool/Cargo.toml --release
make: cargo: No such file or directory
make: *** [src/board/system76/common/common.mk:51: flash_internal] Error 127
```
Users who didn't have `cargo` installed before (like me) will probably run into the same issue.

I'd like to modify this PR to address this issue. I guess that sourcing `~/.bashrc` at the end of `./script/deps.sh` script isn't the proper solution (though it would fix it). What's a good solution that I can include in this PR?

crawfxrd:
`${HOME}/.cargo/env` _is_ sourced by `deps.sh`, immediately after installing rustup.

The only way it won't work is if the user installed rustup previously on their own without modifying `PATH`.

larsblumberg:
I didn't have `rustup` installed before.

Yes, `./scripts/deps.sh` does source `$HOME/.cargo/env` but please note these environment variables are not visible to the outer shell which is invoking `./scripts/deps.sh` and `make flash_internal`.

Since `make flash_internal` does not source the cargo env vars, the build chain will fail if run for the first time on a user's machine like mine. I see at least 3 alternate solutions to this problem.

A) Ask the user the source `$HOME/.cargo/env` (I think that's a poor solution)
B) Ask the user to restart the shell after running `./scripts/deps.sh` (Also a poor solution to my eyes)
C) Have `make flash_install` do the sourcing as part of its build chain (To me that's the better solution).

crawfxrd:
I would rather have the user restart the session or source `~/.cargo/env` themselves because it's a one-time operation: It never needs to be done again.

`cargo` is something that should simply be available, like `sdcc`, without needing to modify the Makefile.

larsblumberg:
I understand! To inform the user about the necessary `source $HOME/.cargo/env` step, would you agree with extending the installation instructions as I proposed in https://github.com/system76/ec/pull/184?

larsblumberg:
Closing in favor of #184

PR #188, opened 2021-04-29
Port Thelio Io to System76 EC

jackpot51:
This includes changes for launch nelson and is therefore still a draft until ectool is cleaned up


jackpot51:
thelio-io-2 now uses QMK

PR #190, opened 2021-05-01
Smooth fan speed changes

curiousercreative:
I wrote this for my new galp5 to address a common complaint of the fans jumping from 0 to 100 and back abruptly. The configuration value of `SMOOTH_FANS_UP` and `SMOOTH_FANS_DOWN` sets the number of fan speed event loops that a 0 - 100 speed ramp should be smoothed across. Effectively, this determines the maximum fan speed adjustment for each event loop to prevent audible stepping. A high number yields a slower and smoother ramp while a low number yields a fast response.

This PR adds support for new build flags:
- `SMOOTH_FANS`  boolean, defaults to `1` for enabled, set to `0` for disabling
- `SMOOTH_FANS_UP` sets the fan smoothing (as described above) for the fan speed increases. Defaults `40`, about 11 seconds
- `SMOOTH_FANS_DOWN` sets the fan smoothing (as described above) for the fan speed decreases. Defaults to `100`, about 25 seconds
- `SMOOTH_FANS_MIN` sets the minimum fan speed (in percent) that should be smoothed. Defaults to 0 (smooth all fan speed changes), but set to 25 for oryp6/7

Attempts to fix https://github.com/system76/firmware-open/issues/149, see prior PR for more history: https://github.com/system76/ec/pull/139


curiousercreative:
@jackpot51 new PR for fan smoothing here. By default this will:
- interpolate fan speed from curve points
- shorter interval for fan speed updates
- smooth a fan speed change of 0 (off) to 100 (full blast) over (approximately) a 10 second window. For my galp5, that translates to about 600RPM change per second or 150RPM per step (4 steps per second)

With build flag `SMOOTH_FANS=0` disables and preserves all current fan speed behavior.

curiousercreative:
I'm running a build from this on a dGPU galp5 and all is working as expected.

src/board/system76/common/peci.c
```diff
@@ -30,7 +30,7 @@ static uint8_t FAN_COOLDOWN[BOARD_COOLDOWN] = { 0 };
 bool peci_on = false;
 int16_t peci_temp = 0;
 
-#define PECI_TEMP(X) (((int16_t)(X)) << 6)
+#define PECI_TEMP(X) ((int16_t)(X))
```
curiousercreative, line 33: @jackpot51 this and line 159 fixes an existing bug with interpolation.

MilesBHuff:
@curiousercreative This is [still](https://github.com/system76/ec/pull/139#issuecomment-822985888) breaking CPU temperature detection on the oryp7:
image

curiousercreative:
@MilesBHuff I only have a galp5 and don't observe such behavior. Hopefully S76 can reproduce and debug, but you may be able to help with the following:
- try running [this script](https://github.com/system76/ec/blob/master/scripts/power.sh) which outputs CSV that you can share
- try running `./scripts/ectool.sh console` from your ec repo folder and that should output the exact values the EC is reading for CPU/PECI and GPU temperature readings as well as the fan speeds it sets in response

MilesBHuff:
Without #190:
- `power.sh`: `00:16:00        0.00    9.9     45.0    90.0    59.0    92      100     11.9    42      4452    4594    4589    4607    4584    4581    4600    4189    3246    3288    4472    4313    4664    46234593     4527`
- `ectool.sh`: `DGPU temp=44` `PECI temp=3185` `BAT 12805 mV 0 mA`
With #190:
- `power.sh`: (tbd)
- `ectool.sh`: (tbd)

curiousercreative:
@MilesBHuff you know, I actually observe the same behavior as you, but I am able to obtain the PECI temperature via coretemp rather than system76_acpi
Screenshot from 2021-05-04 16-53-29

curiousercreative:
@MilesBHuff I might play with this in a day or two, but this is my guess as to where we need to update the interpretation of sensor readings: https://github.com/pop-os/system76-acpi-dkms/blob/master/system76_acpi.c#L503-L542

MilesBHuff:
> I am able to obtain the PECI temperature via coretemp rather than system76_acpi.
Yep!  That's what I'm using right now, as well.  Fun fact:  coretemp has always been different from system76_acpi, so this may be a long-standing bug that was just exacerbated by this PR.

curiousercreative:
@jackpot51 of course, edit freely for any reason :)

jackpot51:
@curiousercreative it should be a pretty simple fix, modify this line to not shift anymore: https://github.com/system76/ec/blob/master/src/board/system76/common/acpi.c#L97

curiousercreative:
@jackpot51 brilliant, works like a charm! My GPU temp as reported by ACPI is still 0C if I'm in "compute" graphics mode, but otherwise did the trick! Also rebased with the latest master.

MilesBHuff:
@curiousercreative @jackpot51 Confirmed fixed on my oryp7!  :D
image

MilesBHuff:
@curiousercreative Confirmed GPU temperature is broken with `compute` graphics:
image
However, this could be entirely unrelated to this change.  We need to test with beforehand.

curiousercreative:
Behavior is the same previously, not a regression.

On May 7, 2021 4:22:10 PM EDT, Miles B Huff ***@***.***> wrote:
***@***.*** Confirmed GPU temperature is broken with `compute`
|graphics:
|image
|However, this could be entirely unrelated to this change.  We need to
|test with beforehand.
|
|-- 
|You are receiving this because you were mentioned.
|Reply to this email directly or view it on GitHub:
|https://github.com/system76/ec/pull/190#issuecomment-834750085

Winston Hoy
(202)630-7559
Curiouser Creative Studio
Video, Photo, Web, Motion, Design

MilesBHuff:
> Behavior is the same previously, not a regression

Oh, that's great news!  I was literally _just_ about to hit "enter" to flash my EC with the old code to compare.

curiousercreative:
You're welcome to confirm, I think there is an issue open on firmware-open repo.

On May 7, 2021 4:26:24 PM EDT, Miles B Huff ***@***.***> wrote:
***@***.*** 
|> Behavior is the same previously, not a regression
|
|Oh, that's great news!  I was literally _just_ about to hit "enter" to
|flash my EC with the old code to compare.
|
|-- 
|You are receiving this because you were mentioned.
|Reply to this email directly or view it on GitHub:
|https://github.com/system76/ec/pull/190#issuecomment-834752383

Winston Hoy
(202)630-7559
Curiouser Creative Studio
Video, Photo, Web, Motion, Design

MilesBHuff:
I am unable to find such an issue there, despite looking at all open or closed issues with "GPU" or "compute" in their name in https://github.com/system76/firmware-open.  A websearch for "system76 github gpu temperature is 0 with compute mode" also turned up nothing.
Can you create an issue there, and link it?

jackpot51:
https://github.com/system76/firmware-open/issues/189

MilesBHuff:
Thanks, @jackpot51.  I didn't think that was the issue, since there was no mention of "compute" graphics.  I'll go ahead and post my screenshot there.

jackpot51:
This looks good to me now. I want QA to take a deep look at this, and then we will release it along with fan syncing

leviport:
I'm currently running this on an oryp6. I'm noticing some strange behavior if I stress the CPU for small amounts of time. It's like the ramp-up sequence initiates, but doesn't stop itself if the CPU load drops back down to idle levels and cools off. 

To reproduce:
- Install `s-tui` and `stress-ng`
- In `s-tui`, arrow down to "Stress" and hit enter. All cores will be stressed and temps will rise.
- After about 5 seconds, arrow back up to "Monitor" and hit enter. The CPU will cool back off, but the fan will continue ramping up, even though the CPU temps have already dropped.

If I only stress it for about 3 seconds, it seems like the fans only go up to about 50% before spinning back down, but the CPU has already cooled off by the time it gets there.

Is there any way to "short circuit" the sequence if temps have already dropped off? Having the fans rev up and back down for brief temp spikes might be kind of annoying.

Also, would we want the same kind of short circuit thing to happen in the other direction? Like if the CPU is hot and the fans are at full, then the CPU only cools off for a little bit, then gets hot again, we wouldn't want it to smooth down, then have to smooth back up, right?

curiousercreative:
@leviport my guess is heatup about cooldown is the culprit. What are your values there? Can you try shortening them to see what effect it has

curiousercreative:
@leviport also, is the behavior not the same prior to fan smoothing or perhaps even more pronounced? If you're using the default heatup and cooldown values (5 and 20), then you should see about 1.25 seconds of lag in the start of a ramp up and about 5 seconds of a lag in reversing a ramp up to ramp down. Without fan smoothing, you should observe a 5 second lag in fans spinning up sharply and a 20 second lag in fan spinning down sharply.

jackpot51:
@curiousercreative do you use different HEATUP/COOLDOWN values, and if so, do you want to try making those the defaults?

leviport:
I do have a fair bit on my plate at the moment, but I should be able to experiment with different heatup/cooldown values soon.

I did want to report something else I noticed while running this EC though. The very slowest bit of smoothing should probably be eliminated altogether, as it makes kind of a buzzing sound when spinning down (presumably from 25% down to 0%). This would probably happen on most oryp6's and oryp7's since we had to tweak the fan curves to keep it from going below 25% in this PR: https://github.com/system76/ec/pull/167.

curiousercreative:
@leviport OK, that might be a simple enough use case to support. Let me see what I can do there

curiousercreative:
@leviport I've added what I hope will solve for your fan buzz. It's working as expected on my galp5, but I don't have an oryp6/7 to verify the fan buzz is any better.

jacobgkau:
I was previously observing the same thing as Levi on galp5, where it seemed like a ramp-up would continue too long after the machine's already cool (or like the ramp-up couldn't be interrupted once it had started.) After cutting the `DBOARD_COOLDOWN` in half from 20 to 10, I'm finding the behavior more desirable-- if I start a stress test, wait for the fans to start ramping up, then kill the stress test, the fans go back down much quicker.

Perhaps we won't need as aggressive of a `DBOARD_COOLDOWN` now that smoothing is implemented, since the actual process of starting/stopping the fans is now less abrupt?

src/board/system76/galp5/board.mk
```diff
@@ -36,23 +36,19 @@ CFLAGS+=\
...
 # Custom fan curve
-CFLAGS+=-DBOARD_HEATUP=5
-CFLAGS+=-DBOARD_COOLDOWN=20
 CFLAGS+=-DBOARD_FAN_POINTS="\
-	FAN_POINT(70, 40), \
-	FAN_POINT(75, 60), \
-	FAN_POINT(80, 75), \
-	FAN_POINT(85, 90), \
-	FAN_POINT(90, 100) \
+	FAN_POINT(70, 25), \
```
curiousercreative, line 40: This is the only fan curve I updated. these are the values I run and they correspond to the system76-power profiles for battery, balanced and performance.

curiousercreative:
@jackpot51 per your suggestion, I'm defaulting to the values that I've found to work well. These are somewhat in contradiction to @jacobgkau and @leviport suggestions to respond more instantly to changes in temperatures, so we may want to sample a larger user pool.

General notes: 
- I caught a work email being used in commits and I've rebased all commits. Sorry for the pain this may cause anyone that has work based on this (@MilesBHuff).
- I've removed any device specific configuration for heatup and cooldown as the smoothed experience is so radically different from before. We can always reintroduce device specific values after the globals are evaluated.
- `SMOOTH_FANS` config is now only a boolean for disabling smoothing. Use `SMOOTH_FANS_UP` and `SMOOTH_FANS_DOWN` to modify the strength of smoothing in either direction.

lucserre:
Just curious, do I have to manually apply this or will it get pushed soon?

MilesBHuff:
Sorry for disappearing!  I've had a lot on my plate lately.

@leviport 
> The very slowest bit of smoothing should probably be eliminated altogether, as it makes kind of a buzzing sound when spinning down (presumably from 25% down to 0%). This would probably happen on most oryp6's and oryp7's since we had to tweak the fan curves to keep it from going below 25% in this PR: #167.

I just flashed the latest on an oryp7, and still hear the buzzing sometimes.  To be fair, I think it's inevitable when spinning down -- on its way to 0%, the fan is always going to have to pass through 24% - 1% in order to reach 0%.  But it _is_ reduced, relative to before.

The latest definitely still works, though!  :)

src/board/system76/common/fan.c
```diff
@@ -105,3 +114,35 @@ uint8_t fan_cooldown(const struct Fan * fan, uint8_t duty) __reentrant {
 
     return highest;
 }
+
+uint8_t fan_smooth(uint8_t last_duty, uint8_t duty) __reentrant {
+  uint8_t next_duty = duty;
```
crawfxrd, line 119: Please indent using 4 columns.

curiousercreative: @crawfxrd sorry about that, I must have coded most of this prior to installing an editorconfig package for my text editor :/ I'll fix this and others up

src/board/system76/common/fan.c
```diff
@@ -4,14 +4,19 @@
 #include <common/debug.h>
 #include <ec/pwm.h>
 
-bool fan_max = false;
+#if SMOOTH_FANS != 0
+    const uint8_t max_jump_up = (MAX_FAN_SPEED - MIN_FAN_SPEED) / (uint8_t) SMOOTH_FANS_UP;
```
crawfxrd, line 8: These can be `static`.

curiousercreative: @crawfxrd is that in place of `const` keyword?

crawfxrd: In addition to it: `static const`.

crawfxrd (approved):

src/board/system76/galp5/board.mk
jacobgkau: You're defining multiple speeds for some of the temperatures here. It seems like the speed is changing around 80, so I take it the highest speed defined for a temperature (or the last point defined for a temperature) takes precedence?

This curve is much quieter, but that's probably more to do with the speeds being nearly cut in half than the smoothing. Running a simple stress test (`stress -c 4 -i 4 -m 4 -d 4`), I'm getting temperatures hovering just above 80 degrees with the new curve, while the old curve hovered a few degrees lower in the high 70's. With the new curve, switching to battery mode brings the temperature below 70 almost immediately, with the fans spinning down soon after. Performance mode sits in the low 90's, keeping the fans at 100%.

Overall, I think this curve will result in a quieter experience for most users. Some users might have a concern about the slightly increased temperatures, but I'd expect the complaints about that to be less than the current complaints about noise levels. galp5's plastic bottom panel won't transmit the heat to the user's lap as much as the metal panel on the lemp10.

Even with this curve, I'm still able to create the behavior (especially in performance mode) where I start a stress test, wait for the fans to start spinning up, stop the stress test, then hear the fans continue to spin up well after the temperature's dropped below 60. I counted six seconds between when the temperature began to drop and when the fans began to drop. With `BOARD_COOLDOWN` and `BOARD_DGPU_COOLDOWN` lowered from 20 to 10, I'm not able to recreate that so easily, as the fans drop down to 40% very soon after the temperatures start falling.

I also see that this PR was what increased the common values for `BOARD_COOLDOWN` and `BOARD_DGPU_COOLDOWN` to 20 in the first place, although I assume it was done because most of the models were overriding with a cooldown of 20. In any case, I would still recommend bringing those down to 10, at least for galp5 where the curve points are so far apart from each other in temperature (making bouncing between points less likely.)

Happy to have @leviport or @bflanagin take a second look on our lab unit, this still feels fairly subjective even when I'm watching the numbers.

MilesBHuff: `10` is also the ideal value for the Oryp7 (and by extension, the Oryp6), as I mentioned a few weeks back.  I know I don't work at System76, but here's one random Internet person's support for making that value default.  :)

curiousercreative:
@crawfxrd sorry, I rebased on master and it dismissed your reviewed :/ I'll let you all rebase on master or merge master in going forward I think.

src/board/system76/galp5/board.mk
curiousercreative: @jacobgkau thanks for the detailed report! I wouldn't be upset if S76 modified default values of course. The effective cooldown time we currently default to is about 5 seconds nominally (6 as you observed though) which is still down considerably from the 10 that we previously defaulted to. It's a good point that the smoothing will take another 10 seconds or so bring the fans all the way to zero, but I was more concerned with bounciness of the fans than the race to zero after load. Hopefully these values will smooth short spikes in load entirely and reduce bounciness for constant loads. There's infinite variability in workloads of course, but I've singled out what I believe are the two most common scenarios and announces associated with them:
1. short spike in load should not result in short spike in fan speeds
2. constant elevated load should not result in fan speeds bouncing noticeably

curiousercreative: @jacobgkau and in response to the multiple fan points, I think you understand correctly the intention to create a pivot point in the fan curve. In theory, we can set just two fan points, a fan speed floor and ceiling and all values in between would interpolate, but to better match with the S76 power profiles, I've added these values

crawfxrd (approved):

src/board/system76/galp5/board.mk
leviport: I also have to agree that I like a cooldown time of `10` on my oryp6. `20` just feels like it goes way too long after a load spike, and the CPU temps have long since dropped off by the time the fan starts smoothing downwards.

@curiousercreative If you want to push that commit, I think this PR is just about ready to be approved.

jacobgkau: I went ahead and committed that change (we're excited to get these improvements in!), I'll make sure that took effect and then we'll approve this.

jacobgkau (approved):

leviport (approved):

MilesBHuff:
Any updates?  :)

crawfxrd (approved):

jackpot51 (approved):

MilesBHuff:
(dummy comment so people can 🎉 react)

MilesBHuff:
Do you guys have an rough guess of when this might be released in an official BIOS update?

crawfxrd:
After the gaze16 work is done.

jackpot51:
We will likely release production firmware for these new products including these changes at the same time we release these changes for prior products, to simplify the updates and keep all products in sync.

ZeddieXX:
> 
> 
> After the gaze16 work is done.
> 
> 
> We will likely release production firmware for these new products including these changes at the same time we release these changes for prior products, to simplify the updates and keep all products in sync.

I see the  gaze16 is now available to configure.  Does this mean the firmware will be released soon for all other laptops?  Would love to see this in the production firmware as an official update.

Thanks!

ZeddieXX:
> > After the gaze16 work is done.
> > We will likely release production firmware for these new products including these changes at the same time we release these changes for prior products, to simplify the updates and keep all products in sync.
> 
> I see the gaze16 is now available to configure. Does this mean the firmware will be released soon for all other laptops? Would love to see this in the production firmware as an official update.
> 
> Thanks!

@jackpot51 is the gaze16 already out?  If so, is a new firmware release with these merges be officially released and sent out as an update via the built-in firmware updater in Pop!_OS?

leviport:
@ZeddieXX  https://github.com/system76/firmware/pull/321

jacobgkau:
@ZeddieXX You will receive a notification when the firmware is released. Testing is still ongoing, there is no guaranteed timeline.

ZeddieXX:
> 
> 
> @ZeddieXX You will receive a notification when the firmware is released. Testing is still ongoing, there is no guaranteed timeline.
> 
> If you don't want to wait, you are welcome to flash updated firmware yourself by installing [Rust nightly](https://rustup.rs/) and then running these commands:
> 
> ```
> git clone https://github.com/system76/firmware-open
> cd firmware-open
> ./scripts/update.sh
> ./scripts/deps.sh
> ./scripts/build.sh <your-model>
> ./scripts/flash.sh <your-model>
> ```

Thank you for this.  I am not an expert at this, being a Windows user.  The readme got to the build, but not how to flash.  And even then I needed rust installed apparently.  This is why I was asking about the released version, but if I can download, compile, and flash it myself, I'll do it.

When the official firmware comes out, will I still get notifications since I'll be using a non-release firmware?

jacobgkau:
> When the official firmware comes out, will I still get notifications since I'll be using a non-release firmware?

Once you flash a custom version, you'll actually start receiving notifications immediately, as the released version would be an available "update" from your custom version as far as the firmware manager is concerned. You can switch back to released firmware at any time by following the normal update process, before or after the new firmware is released.

I still don't have an estimate on the new firmware release, but there was a model-specific issue (unrelated to the fan changes) that was discovered during testing, which is on the tail end of being resolved.

ZeddieXX:
> 
> 
> > When the official firmware comes out, will I still get notifications since I'll be using a non-release firmware?
> 
> Once you flash a custom version, you'll actually start receiving notifications immediately, as the released version would be an available "update" from your custom version as far as the firmware manager is concerned. You can switch back to released firmware at any time by following the normal update process, before or after the new firmware is released. (You will see the new version show up in the firmware manager's changelog once it's released.)
> 
> I still don't have an estimate on the new firmware release, but there was a model-specific issue (unrelated to the fan changes) that was discovered during testing, which is on the tail end of being resolved.

Got it flashed!  Immediate difference in fan and noise, so thank you for this!

Issue #187, opened 2021-04-27
Holding key during S3 resume results in EC timeout


crawfxrd:
### Steps to reproduce

1. Suspend the system with `systemctl suspend`
2. Press and hold any key

### Expected results

System wakes from S3 without excess delay and without errors/warnings in dmesg related to EC functionality.

### Actual results

board: system76/oryp5
version: 2021-01-26_6213d00-dirty (12a9ac8d84)

```
[20772.673749] ACPI: Waking up from system sleep state S3
[20772.673779] ACPI Debug:  "LID: _PSW: 0000000000000000"
[20776.703242] ACPI Error: AE_TIME, Returned by Handler for [EmbeddedControl] (20201113/evregion-294)
[20776.703245] ACPI Error: Timeout from EC hardware or EC device driver (20201113/evregion-304)

[20776.703250] No Local Variables are initialized for Method [_PSW]

[20776.703252] Initialized Arguments for Method [_PSW]:  (1 arguments defined for method invocation)
[20776.703252]   Arg0:   000000001259483c <Obj>           Integer 0000000000000000

[20776.703257] ACPI Error: Aborting method \_SB.LID0._PSW due to previous error (AE_TIME) (20201113/psparse-529)
[20776.703262] ACPI: _PSW execution failed
[20776.703351] ACPI Debug:  "EC: WAK: 0000000000000003"
[20779.757805] ACPI Error: AE_TIME, Returned by Handler for [EmbeddedControl] (20201113/evregion-294)
[20779.757807] ACPI Error: Timeout from EC hardware or EC device driver (20201113/evregion-304)

[20779.757812] No Local Variables are initialized for Method [WAK_]

[20779.757813] Initialized Arguments for Method [WAK_]:  (1 arguments defined for method invocation)
[20779.757814]   Arg0:   000000001259483c <Obj>           Integer 0000000000000003

[20779.757818] ACPI Error: Aborting method \_SB.PCI0.LPCB.EC0.WAK due to previous error (AE_TIME) (20201113/psparse-529)
[20779.757821] ACPI Error: Aborting method \_SB.MWAK due to previous error (AE_TIME) (20201113/psparse-529)
[20779.757824] ACPI Error: Aborting method \_WAK due to previous error (AE_TIME) (20201113/psparse-529)
[20779.757828] ACPI Error: AE_TIME, While executing method \_WAK (20201113/hwesleep-47)
[20779.757957] ACPI: EC: interrupt unblocked
[...]

PR #194, opened 2021-05-13
kbc: Reduce timeout from 10ms to 1ms

crawfxrd:
Fix ACPI timeout and delay on S3 resume when holding a key.

```
[ 3368.051132] ACPI: Waking up from system sleep state S3
[ 3369.136271] ACPI: EC: interrupt unblocked
[ 3370.249561] ACPI: EC: event unblocked
```

Tested on: oryp5, oryp6

While a timeout of 5ms was sufficient to fix the ACPI error, a value greater than 1ms (tested with 1.5ms and 2ms) still results in the excessive delay (~10s).

Resolves: #187

May also address system76/firmware-open#156

### DIFF
--- a/src/board/system76/common/kbc.c
+++ b/src/board/system76/common/kbc.c
@@ -22,7 +22,7 @@ void kbc_init(void) {
     *(KBC.status) = BIT(4);
 }
 
-#define KBC_TIMEOUT 10000
+#define KBC_TIMEOUT 1000
 
 // Enable first port - TODO
 static bool kbc_first = false;


jackpot51 (dismissed):

jackpot51 (approved):

jackpot51:
I wonder if this also improves slow boot times

crawfxrd:
It might. @jhorsm reported ACPI timeouts on darp6. I'm not able to reproduce the slow prompt typing on oryp6, though.

leviport (approved):
I tried this EC on my darp5 and it doesn't seem to fix the slow encryption prompt typing issue. However, I'm also not seeing any regressions.

leviport:
Scratch that, it seems to have started working after I rebooted a few times? I'll try it on my oryp6 next to see if it makes a difference on that as well.

leviport:
The first couple boots on the oryp6 had a much less pronounced delay, but I still noticed it a bit, then it went away and seems completely fixed. Since it seemed kinda intermittent, I tried my darp5 a few more times and saw the delay once.

So even though it's still happening, at least this makes it less frequent? Either way, I know this PR wasn't aimed at that problem, and any improvement on that is pure gravy. I think it's good to merge.

PR #198, opened 2021-05-18
Add interrupt controller registers

crawfxrd:
Definitions for the INTC and WUC modules.

- IT8587E has INTC groups 0-18
- IT5570E has INTC groups 0-21
- WUC groups  1, 3, 4 have an extra register for enabling interrupts to INTC
- WUC does not have a group 5

External interrupts are enabled, but as all sources are masked, INT1# is never asserted.

### DIFF
--- a/src/arch/8051/arch.c
+++ b/src/arch/8051/arch.c
@@ -11,6 +11,9 @@ void arch_init(void) {
 
     time_init();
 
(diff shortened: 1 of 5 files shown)


crawfxrd:
Included in #204, where WUC/INTC is used.

Issue #195, opened 2021-05-13
galp3-c: Touchpad not always restored on S3 resume


crawfxrd:
This blocks the Open Firmware release for galp3-c and galp4.

### Affects

- galp3-c
- galp4

### Steps to reproduce

- Perform S3 suspend
- Wake system
- Check touchpad functionality
- Repeat

To ease reproduction of this issue, I am using a modified `suspender.sh` script from the private testing repo. It checks for the existence of`/sys/devices/platform/i8042/serio1/input` instead of running a certain number of times.

### Expected behavior

The touchpad functions on every resume.

### Actual behavior

Consistently, but after an arbitrary amount of suspends, the touchpad no longer functions.

After enabling `TRACE` debugging, I see that all `kbc mouse` logs stop occurring once the device is lost.

In the case of the galp3-c, functionality does not seem to be restored until a reboot. (Although I was not particularly interested in seeing it would return after more suspend cycles, so only attempted more suspend cycles once.)

crawfxrd:
With `i8042.debug=1` on Linux, you can see that IRQ12 stops happening.

- [Good](https://github.com/system76/ec/files/6495252/s3-good.txt): Touchpad restored on resume
- [Bad](https://github.com/system76/ec/files/6495251/s3-bad.txt): Touchpad not restored, no IRQ12

crawfxrd:
The mouse reset (0xDD 0xFF) never completes: `TDS` isn't set after 10k (`PS2_TIMEOUT`) reads, but neither is `TOER`. It's normally complete on the first read.

I see more ACPI timeouts related to this as well.

```
[  403.492994] i8042: [100527] d4 -> i8042 (command)
[  403.493581] i8042: [100527] ff -> i8042 (parameter)
[  403.574765] usb 1-7: reset high-speed USB device number 3 using xhci_hcd
[  403.583216] ata1: SATA link down (SStatus 4 SControl 300)
[  403.749599] OOM killer enabled.
[  403.749603] Restarting tasks ... done.
[  403.753510] video LNXVIDEO:00: Restoring backlight state
[  403.774779] PM: suspend exit
[  404.514746] i8042: [100783] d4 -> i8042 (command)
[  405.138280] i8042: [100938]      -- i8042 (wait write timeout)
[  405.991647] ACPI Error: AE_TIME, Returned by Handler for [EmbeddedControl] (20201113/evregion-294)
[...]

PR #201, opened 2021-05-24
Make touchpad access non-blocking

jackpot51:
When using I2C HID, the touchpad stops responding over PS/2. We should ensure this minimally affects other parts of the EC firmware

### DIFF
--- a/src/board/system76/common/kbc.c
+++ b/src/board/system76/common/kbc.c
@@ -24,10 +24,12 @@ void kbc_init(void) {
 
 #define KBC_TIMEOUT 1000
 
-// Enable first port - TODO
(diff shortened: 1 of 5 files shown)


jackpot51:
This will need to be tested on the galp3-c or galp4 which have no I2C HID.

crawfxrd:
galp3-c touchpad still works with this change. Trying to reproduce #195 as well.

jacobgkau:
Touchpad also still works on galp4.

crawfxrd (approved):

jacobgkau (approved):
Fixes the input lag at boot on oryp7.

PR #204, opened 2021-05-27
Wake from `LAN_WAKEUP#` interrupt

crawfxrd:
Replace polling logic for Wake-on-LAN with an ISR for `LAN_WAKEUP#`.

Tested waking from S5 and S3 on galp3-c and gaze15.

Waking from S3 apparently doesn't work on oryp6, but that is not caused by this change.

For how to test, see #89.

### DIFF
--- a/src/arch/8051/arch.c
+++ b/src/arch/8051/arch.c
@@ -11,6 +11,9 @@ void arch_init(void) {
 
     time_init();
 
+    // Enable external interrupts for INTC
+    EX1 = 1;
+
     // Enable interrupts
     EA = 1;
 }
--- a/src/board/system76/addw1/interrupts.c
+++ b/src/board/system76/addw1/interrupts.c
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+#include <board/interrupts.h>
+#include <ec/intc.h>
+#include <ec/wuc.h>
+
+void interrupts_init(void) {
+}
+
+void external_1(void) __interrupt(2) {
+}
--- a/src/board/system76/addw1/interrupts.c
+++ b/src/board/system76/addw1/interrupts.c
@@ -1,11 +1,24 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
 #include <board/interrupts.h>
+#include <board/power.h>
+#include <common/macro.h>
 #include <ec/intc.h>
 #include <ec/wuc.h>
 
 void interrupts_init(void) {
+    // INT92: LAN_WAKEUP#
+    WUEMR8 = BIT(4);
+    WUESR8 = BIT(4);
+    ISR11 = BIT(4);
+    IER11 = BIT(4);
 }
 
 void external_1(void) __interrupt(2) {
+    if (ISR11 & BIT(4)) {
+        // INT92: LAN_WAKEUP#
+        power_handle_lan_wakeup();
+        ISR11 = BIT(4);
+        WUESR8 = BIT(4);
+    }
 }
--- a/src/board/system76/addw2/interrupts.c
+++ b/src/board/system76/addw2/interrupts.c
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+#include <board/interrupts.h>
+#include <ec/intc.h>
+#include <ec/wuc.h>
+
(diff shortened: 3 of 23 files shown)


jacobgkau:
Waking with a magic packet still works on galp5. I noticed that on WoL boots, suspend/resume doesn't work (wakes back up immediately), but that's not a regression. On non-WoL boots, it can't be woken from suspend using WoL, but that's also not a regression.

On both galp5 and darp7, I'm getting very bad keyboard lag with this branch (similar to what https://github.com/system76/ec/pull/201 fixed, except it doesn't go away after boot.) On darp7, the machine's not powering off with this firmware-- if I select `Power Off` from the top-right menu (or even just press the power button at the firmware menu), it powers off and then powers right back on.

crawfxrd:
I somehow missed that darp7 uses GPH4 for `LAN_WAKEUP#` instead of GPB2. That's fixed.

galp5 does use GPB2, so I'm not sure what's wrong there. Can you get dmesg and EC console output from it?

jacobgkau:
WoL is no longer working on galp5 after the latest changes. When I set `Wake-on` to `g` using ethtool (confirming that it worked by checking with ethtool immediately afterwards), then power off the machine (but keep it plugged in), upon powering back on using the power button, `Wake-on` is set back to `d`.

For the keyboard issue, here are some outputs from galp5:
[dmesg-laggy-keyboard-1.txt](https://github.com/system76/ec/files/6561475/dmesg-laggy-keyboard-1.txt)
[ec-console-1.txt](https://github.com/system76/ec/files/6561477/ec-console-1.txt)
[dmesg-laggy-keyboard-2.txt](https://github.com/system76/ec/files/6561476/dmesg-laggy-keyboard-2.txt)
[ec-console-2.txt](https://github.com/system76/ec/files/6561478/ec-console-2.txt)

Interestingly, while running the ectool console, if I try to type "system76", I see entries in the console for every character, but only some of them are actually input:
Screenshot from 2021-05-28 07-03-50

WoL does work on the darp7 and its power-off issue is now fixed, but it's also experiencing the keyboard issue.

crawfxrd:
> upon powering back on using the power button, `Wake-on` is set back to `d`.

This is expected: it gets disabled on every boot.

I've dropped the GPIO change, so WoL should work on galp5 again.

Can you confirm that keyboard works correctly on 38a83c62bbe ("Add empty, per-board init function for INTC")?

jacobgkau:
Confirmed that galp5 WoL works on d60ddc0. I realized that I may have been using the wrong MAC address for galp5 earlier, and got thrown off by seeing the `Wake-on` setting reset on boot (I see now that does happen even when it's working.) I tried to go back to the version with the GPIO change, but you force-pushed over that commit; if you put that back, I can try it again. Sorry about that.

For the keyboard issue, it still happens on d60ddc0, but the issue does not happen on 38a83c6. The keyboard works correctly on that commit.

jacobgkau:
WoL works fine on 2b1f2c9, the only remaining issue is the keyboard.

jacobgkau:
I'm getting no power or charging LED after flashing dd9ad84.

crawfxrd:
Well, I kind of expected that...It's stuck in the ISR and not running any other code.

Does that happens on both darp7 and galp5?

Are you able to configure the external console for debugging?

jacobgkau:
Just to summarize, after dropping commit fe725a9, enabling debugging, and reflashing, I see `INT159` being spammed in the console when it stops powering on.

crawfxrd:
> (Always Enabled) | PLL Frequency Change Event Interrupt

Okay. So this isn't and can't be masked?

And it's only happening on the TGL boards because those have to use eSPI which requires changing `PLLFREQ`.

crawfxrd:
I'm not sure what, if anything, needs to be done for the `PLLFREQ` change. So for now I just ack it.

(I'm not sure _when_ the interrupt occurs: When the `PLLFREQ` register changes? After the sleep when it takes effect? But it could probably at least do the delay waiting for PLL to stabilize?)

jacobgkau (approved):
After the latest changes, galp5/darp7 no longer have a laggy keyboard and WoL still works.

crawfxrd:
Going to make this WIP while I work on substantial changes to the interrupt handling I added.

PR #205, opened 2021-06-01
Reduce copying of keyboard layouts

jackpot51:
This makes all system76 boards which share the same keyboard layout use the same code, moved to src/keyboard/system76/*


crawfxrd (approved):

leviport:
This doesn't fix the [M key problem](https://github.com/system76/firmware/pull/300#issuecomment-811400245) on my darp5, but it also doesn't seem to cause any new problems.

crawfxrd:
Paths in `doc/keyboard-layout-customization.md` should be updated.

jackpot51:
Documentation updated, please take a look

crawfxrd (approved):

PR #206, opened 2021-06-03
Define GPCRn values in EC header

crawfxrd:
Define GPCRn values once per EC, since they are not board-specific.


jackpot51 (approved):

jacobgkau (approved):
Still builds and flashes successfully (tested on galp5.)

PR #216, opened 2021-06-09
ec/it*: Remove vendor string from signature

crawfxrd:
Per the specs, a 16-byte signature is stored at 0x40. All proprietary EC ROMs store a space-padded 16-byte vendor string at 0x50, but it isn't required.


jackpot51 (dismissed):

jackpot51:
I'd prefer to take the vendor string out if that is even possible.

crawfxrd:
It's not in the specs, so I think it's only for tools to report the vendor (probably the situation as #127).

jackpot51 (approved):

PR #217, opened 2021-06-10
Merge ITE EC code

crawfxrd:
Most of the code for IT8587E and IT5570E is the same. The primary differences are:

- IT5570E adds new registers
- IT5570E adds support for eSPI

Merge them into a single ITE directory.


src/ec/ite/ec.mk
```diff
@@ -0,0 +1,13 @@
+# SPDX-License-Identifier: GPL-3.0-only
+
+ARCH=8051
+
+# 64 KB is the max without banking
+CODE_SIZE=65536
+
+
```
jackpot51, line 8: There is an extra empty line

jackpot51 (approved):

jacobgkau:
darp7 (5570E) is not powering on after flashing this branch. I have no LED feedback while unplugged, and both LEDs remain solid orange while plugged in.

crawfxrd:
Ah, right. `__EC__` isn't usable for this. I'll have to add a new define to use.

crawfxrd:
Boards using IT5570E now produce identical ROMs if the version is set to the same thing.

Boards using IT8587E are different, due to the empty `ec_read_post_codes()` function and `SCAR0` defines. SDCC has relocated things, and the ROM is 1 byte bigger. This would be simple enough to change for now so it generates identical ROMs as well.

jackpot51 (approved):

jacobgkau (approved):
darp7 (5570E) and galp4 (8587E) both flash, boot, and appear to work normally with this branch.

PR #218, opened 2021-06-10
Move power limit functions to common board code

crawfxrd:
Logic for changing power limit is the same for all boards with GPUs. It is still called from `board_event()` instead of in `power_event()` to maintain the current behavior of checking every main cycle.

### DIFF
--- a/src/board/system76/galp5/board.c
+++ b/src/board/system76/galp5/board.c
@@ -6,7 +6,6 @@
 #include <board/espi.h>
 #include <board/gctrl.h>
 #include <board/gpio.h>
-#include <board/peci.h>
 #include <board/power.h>
 #include <common/debug.h>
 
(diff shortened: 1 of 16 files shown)


src/board/system76/galp5/board.c
```diff
@@ -38,48 +37,9 @@ void board_init(void) {
...
-}
-#else // HAVE_DGPU
-void board_on_ac(bool ac) { /* Fix unused variable */ ac = ac; }
-#endif // HAVE_DGPU
-
 void board_event(void) {
 #if HAVE_DGPU
-    static bool last_power_limit_ac = true;
-    // We don't use power_state because the latency needs to be low
-    if (gpio_get(&CPU_C10_GATE_N)) {
```
jackpot51, line 74: ESPI boards must use CPU_C10_GATE_N, ensure this is handled correctly

crawfxrd: Done

jackpot51 (approved):

PR #224, opened 2021-07-16
Optimize `delay_ms()`

crawfxrd:
Use u8 instead of i16 as a delay can never be negative and should never block for a long period of time.

Use a decrementing loop as SDCC generates more efficient code.


jackpot51 (approved):

jacobgkau (approved):
Tested on darp7, built/flashed successfully, not seeing any side effects.

I did notice that firmware-setup is responding a little slowly (lagging when I push the up/down arrow keys or ignoring some keypresses), but that's also happening with master EC. (It was happening before I merged master into this branch.)

jackpot51 (approved):

Issue #223, opened 2021-07-13
How to define keycode for "scroll lock"?


larsblumberg:
I use the `Scroll Lock` key as the "compose key" to type "special characters" on Pop_OS! (configurable via _Settings > Keyboard > Type Special Characters_).

The firmware currently doesn't define key codes for the scroll lock key as documented here:

https://github.com/system76/ec/blob/c7fafe01038d3adbc9893cd5ed2214f5424563bb/src/common/include/common/keymap.h#L170

I'd like to map the `Fn + Print Screen` key to `Scroll Lock`.

Is there a chance to find and define the corresponding key codes myself or is there more engineering effort involved than just defining the codes?

crawfxrd:
Scroll lock can now be defined with `K_SCROLL_LOCK` in the keymap.

larsblumberg:
Thanks a ton @crawfxrd!

larsblumberg:
@crawfxrd Just wanted to confirm that I can type special characters now (such as German umlauts) using the Scroll Lock key on my lemp10. Thanks again for your changes once more.

PR #225, opened 2021-07-16
Add scroll lock to default keymaps

crawfxrd:
I don't know anything that uses this, so I only checked that xev reported the key as scroll lock.

Resolves: #223


jackpot51 (dismissed):

leviport:
- [ ] darp6 ISO layout (15" 102 key, ISO)
- [ ] darp6 ANSI layout (15" 102 key)
- [ ] bonw14 (15" 102 key with n-key rollover) -- **No lab unit here to test with**
- [ ] galp4 (14" 86 key)
- [ ] lemp10 (14" 83 key)

leviport (changes requested):
I'm not sure we should add Scroll Lock to 14in_83 since Scroll Lock was never printed on lemp9/lemp10's keyboard. The preprod lemp9 we have has it printed, but the ones we actually shipped do not.

crawfxrd:
Dropped from `14in_83`. galp5 also uses it, but I don't think they print the function on the key either.

leviport:
Yep, galp5's keycaps are printed the same as lemp9/10. 

We'll get the bonw14 tested later today. If that looks good, then we'll approve.

jackpot51 (approved):

jacobgkau (approved):
bonw14's working fine as well. On Pop!_OS, I did need to run `xmodmap -e 'add mod3 = Scroll_Lock'` before the scroll lock would actually toggle on and off (before that, I saw the key being pressed, but the toggle wouldn't turn on.) The Insert key and the rest of the keyboard still work as expected.

PR #226, opened 2021-07-19
Use unsigned, shorter data types

crawfxrd:
Ref: #222

```diff
--- build/system76/gaze15/2021-08-02_99af8a3/ec.mem     2021-08-02 12:14:45.885302912 -0600
+++ build/system76/gaze15/2021-08-02_6eff43f/ec.mem     2021-08-02 12:15:02.065163105 -0600
@@ -1,12 +1,12 @@
 Internal RAM layout:
       0 1 2 3 4 5 6 7 8 9 A B C D E F
 0x00:|0|0|0|0|0|0|0|0|a|a|a|a|a|a|b|b|
-0x10:|b|b|b|b|b|b|b|b|b|b|b|b|b|c|c|c|
-0x20:|c|c|c|c|c|c|c|d|e|e|e|e|e|e|e|e|
-0x30:|f|f|g|g|g|g|g|g|g|g|g|g|g|g|g|g|
-0x40:|g|g|g|g|g|g|g|g|g|g|g|g|g|g|g|g|
-0x50:|g|h|i|i|i|i|i|i|i|i|i|i|Q|Q|Q|Q|
-0x60:|Q|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
+0x10:|b|b|b|b|b|b|b|b|b|b|b|b|c|c|c|c|
+0x20:|d|e|e|e|e|e|e|e|e|f|f|f|f|f|f|f|
+0x30:|f|f|f|f|f|f|f|f|f|f|f|f|f|f|f|f|
+0x40:|f|f|f|f|f|f|f|f|g|h|h|h|h|h|h|h|
+0x50:|h|h|h|Q|Q|Q|Q|Q|S|S|S|S|S|S|S|S|
+0x60:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
 0x70:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
 0x80:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
 0x90:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
@@ -18,12 +18,12 @@
 0xf0:|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|S|
 0-3:Reg Banks, T:Bit regs, a-z:Data, B:Bits, Q:Overlay, I:iData, S:Stack, A:Absolute
 
-Stack starts at: 0x61 (sp set to 0x60) with 159 bytes available.
[...]

Issue #222, opened 2021-07-10
Replace use of bare `int` with sized types from `stdint`


crawfxrd:
`int` is a signed 16 bit value. In most cases, they can be unsigned values. In many cases, their use can be an 8 bit value. Either way, sized types from `stdint` should be used.

For example, changing a loop index from `int` to `uint8_t` can reduce the condition to a single `cjne`.

```diff
 00389$:
-;      src/board/system76/common/power.c:463: for (int i = 0; i < 100; i++) {
-       mov     r4,#0x00
+;      src/board/system76/common/power.c:463: for (uint8_t i = 0; i < 100; i++) {
        mov     r5,#0x00
 00195$:
-       clr     c
-       mov     a,r4
-       subb    a,#0x64
-       mov     a,r5
-       xrl     a,#0x80
-       subb    a,#0x80
-       jc      00390$
-       ljmp    00121$
+       cjne    r5,#0x64,00390$
 00390$:
+       jnc     00121$

```

crawfxrd:
For some reason, using `short` (which is part of the `{u,}int16_t` type) may cause SDCC to add 2 unnecessary `mov`s.

https://godbolt.org/z/Ec4zeavPj

crawfxrd:
Related, literals should use `U`/`UL` so SDCC will treat them as unsigned. Can potentially avoid integer promotion and library function calls (or at least directly call the unsigned version if application, e.g., `mod{s,u}int()`).

crawfxrd:
Some unnecessary `mov`s should be fixed by [peephole 302](https://sourceforge.net/p/sdcc/code/HEAD/tree/tags/sdcc-4.1.0/sdcc/src/mcs51/peeph.def#l5040) in SDCC 4.1.0. But this doesn't resolve the ones seen in Compiler Explorer when `short` is used.

PR #228, opened 2021-07-28
Use explicitly sized types from stdint

crawfxrd:
Replace bare `int` types with stdint types. This was done with:

    grep -rwl 'int' src/ | xargs sed -i 's/\<int\>/int16_t/g'
    grep -rwl 'unsigned long' src/ | xargs sed -i 's/\<unsigned long\>/uint32_t/g'
    grep -rwl 'unsigned char' src/ | xargs sed -i 's/\<unsigned char\>/uint8_t/g'

Then reverted for `*main()`, `putchar()`, and `getchar()`.

The Arduino declarations for `parallel_main()` were also corrected to match their definitions.

SDCC does *not* generate the same code in all instances, due to `int` being treated different than `short int`.

Ref: #222 

#226 will be modified to be a follow up that reduces type widths and/or makes them unsigned where practical.


jacobgkau (approved):
Builds & flashes on darp7, not observing any problems with suspend/resume, typing, touchpad, or charging.

jackpot51 (approved):

Issue #197, opened 2021-05-18
Pause/Break not implemented, through printed on the keys for several models


ids1024:
This appears on the "PgUp" and "PgDn" keys at the top right of several models with numpads.

Apparently "Break" on a PC is really `Ctrl + Pause`. Usually it is not a separate key. I haven't tested with proprietary firmware, but I would guess `fn + PgUp` sends `Pause`, and `fn + PgDn` sends `Ctrl` and `Pause`.

Not sure what behavior we really want, and these keys aren't very important, but probably `Pause` should at least be an option in the Configurator.

MilesBHuff:
Can confirm:  I discovered this around the same time you did when I was writing [the Arch Wiki article for the Oryx Pro series](https://wiki.archlinux.org/title/System76_Oryx_Pro#Function_keys).
![image](https://user-images.githubusercontent.com/6566982/169742415-d6cb8ac9-b46c-44ab-a129-ed92ca4036b0.png)
At present, the Fn key seems to fall through on the Pause/Break keys, and we get PgUp or PgDn instead.

PR #229, opened 2021-07-29
Add Pause key

crawfxrd:
On proprietary firmware and in QMK, `Pause` and `Break` are the same thing. `Ctrl` must be held for the OS to register the key as `Break`.

Behavior tested with xev.

Resolves: #197


ids1024:
I guess the Configurator also needs some way to know if the EC firmware is a newer version with support for these. (Just comparing the date from the version would more or less work, I suppose, without a better solution.)

jackpot51 (approved):

jacobgkau:
On serw12 (proprietary firmware), `Pause` and `Break` both show up as keycode 127 (`Pause`). On oryp7 with this branch, `Pause` shows up as keycode 127, but `Break` shows up as keycode 37 (`Control L`).

crawfxrd:
Yeah. Looking at the kudu6, they're both `Pause`, and you have to explicitly hold `LCtrl` to get `Break`. The 14in_86 keyboard (galp3-c/galp4) probably has the same behavior.

And I just checked QMK, it defines `KC_BRK` to `KB_PAUSE`.

So, no `Break`.

jackpot51 (approved):

jacobgkau (approved):
Pause and Break are both showing up with keycode 127 in `xev` as expected, PgUp/PgDn are still working normally when not holding Fn down.

Pause/Break are not showing up in the keyboard configurator, but that doesn't seem like an issue with the EC.

jackpot51 (approved):

PR #230, opened 2021-07-29
Remove `layouts.sh`, and add a `K_NONE` define

ids1024:


crawfxrd:
I wonder if we should change the meaning of `___` to be `KC_TRANSPARENT` instead of `KC_NO` to match QMK.

crawfxrd (approved):

PR #231, opened 2021-07-30
gaze16-3050: Correct charger input current

crawfxrd:
Per the charger for the 3050 variants:

    DC OUTPUT (输出/輸出): 19.5V⎓7.7A 150W

The 3060 charger does list 9.23A.


jacobgkau (approved):
Before the change, stressing the system usually sat at 150W, but if I started stressing in High Performance mode unplugged and then plugged in, I saw a peak as high as 216W. After the change, I'm still seeing peaks as high as 200+ watts when I perform that test.

When powered off, the system drew less than 30W (peaked at 29.7W) before, and still does after the change.

Before the change, I did see one case where the GPU gave an error in gpu_burn when I unplugged the charger; ~~I did not see that after the change~~ I saw that happen again as well after the change.

I don't see any difference in behavior here, but I also don't see any regressions.

jackpot51 (approved):

Issue #233, opened 2021-08-02
darp7: Can power on when unplugged, lid closed


crawfxrd:
## Steps to reproduce:

1. Power off machine
2. Unplug AC adapter
3. Close lid
4. Repeatedly press the power button

## Expected behavior

The machine does not turn on.

## Actual behavior

The machine occasionally boots.

## Notes

Tested on:

- ff53939
- 73c6bd9

`LID_SW#` may be high early in the boot, causing the check to disable the power button to fail. In some cases, `lid_event()` reports "Lid open" (and thus updates `lid_state`, meaning it is also not reliable).

```
[16:06:24] ESGCTRL2: 0 set to 40
[16:06:24] PLLFREQ is 0x01
[16:06:24] PLLFREQ is now 0x07
[16:06:24] ESGCTRL2: 40 set to 0
[16:06:24] ESGCAC2: A set to B

[16:06:24] System76 EC board 'system76/darp7', version '2021-08-02_99af8a3-dirty'
[16:06:24] 00: power_on_ds5
[16:06:24] POWER_STATE_DS5
[16:06:24] 00: SLP_SUS# asserted
[16:06:24] Lid open
```

PR #234, opened 2021-08-03
darp7: Hack-fix power-on issue

crawfxrd:
The power button should be disabled when a machine is unplugged with the lid closed. On darp7, `LID_SW#` can take >200ms at init to reflect the correct state of the lid, so it would regularly fail the check. It was common for the board to still power on.

Ref: #233

### DIFF
--- a/src/board/system76/darp7/board.c
+++ b/src/board/system76/darp7/board.c
@@ -1,10 +1,10 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
+#include <arch/delay.h>
 #include <board/board.h>
 #include <board/espi.h>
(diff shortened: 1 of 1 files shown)


jacobgkau:
Is there any side effect to this (like having to hold down the power button noticeably longer), or is this a fix we'd be able to ship?

crawfxrd:
On AC it should not be noticeable, since the delay will happen when AC power is connected, not when the power button is pressed.

On battery it _may_ be noticeable, since pressing the power button is what turns the EC on in the first place. It will increase the time needed to hold the button from ~100 ms to ~300 ms.

PR #237, opened 2021-08-16
Remove overflow check for unsigned diffs

crawfxrd:
Calculating diffs with *unsigned* integers will always produce the correct result, as long as the diff is less than the type max.

For example, suppose we read 254 and 2 for an 8 bit value:

    2 - 254 = 4

Using this property, simplify diffs on `time_get()` comparisons.


jackpot51 (approved):

jacobgkau (approved):
Builds & flashes on oryp7, fan behavior and LED blinking behavior seem identical to how they were before.

PR #243, opened 2021-09-15
Read cycle count from battery and expose for ACPI

ids1024:


jackpot51 (approved):
This looks pretty good to me

crawfxrd (approved):

Issue #241, opened 2021-09-09
Latest Oryx Pro Firmware Update - Fan Usage/Noise Issue v:2021-07-20_93c2809


barthvader13:
I upgraded my oryx pro firmware to the `2021-07-20_93c2809` version today.

Now the fans start randomly humming and every so often make kind of a low frequency weird chugging click sound. 

I have slack, discord, chrome, postman, sublime, and a single terminal window open with two external monitors plugged in on Nvidia Graphics + High Performance Mode.

comp info:
```
NAME="Pop!_OS"
VERSION="21.04"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 21.04"
VERSION_ID="21.04"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=hirsute
UBUNTU_CODENAME=hirsute
LOGO=distributor-logo-pop-os
```

sensors output:
```
iwlwifi_1-virtual-0
Adapter: Virtual device
temp1:            N/A  

coretemp-isa-0000
Adapter: ISA adapter
Package id 0:  +55.0°C  (high = +100.0°C, crit = +100.0°C)
Core 0:        +54.0°C  (high = +100.0°C, crit = +100.0°C)
Core 1:        +54.0°C  (high = +100.0°C, crit = +100.0°C)
Core 2:        +53.0°C  (high = +100.0°C, crit = +100.0°C)
Core 3:        +53.0°C  (high = +100.0°C, crit = +100.0°C)
Core 4:        +52.0°C  (high = +100.0°C, crit = +100.0°C)
Core 5:        +55.0°C  (high = +100.0°C, crit = +100.0°C)
Core 6:        +54.0°C  (high = +100.0°C, crit = +100.0°C)
Core 7:        +56.0°C  (high = +100.0°C, crit = +100.0°C)

nvme-pci-2700
Adapter: PCI adapter
Composite:    +36.9°C  (low  = -273.1°C, high = +84.8°C)
                       (crit = +84.8°C)
Sensor 1:     +36.9°C  (low  = -273.1°C, high = +65261.8°C)
Sensor 2:     +35.9°C  (low  = -273.1°C, high = +65261.8°C)

pch_cometlake-virtual-0
Adapter: Virtual device
[...]

PR #245, opened 2021-09-29
Adjust fan curve to prevent constant toggling of fan

CourchesneA:
Should fix 
https://github.com/system76/firmware-open/issues/250
and possibly
https://github.com/system76/ec/issues/241

With the fan curve, if the CPU/GPU temp is oscillating between two curve points, the fan will constantly produce annoying sound when they start and stop. This PR adds another curve point to make a smoother transition.

This is only a quick fix, it is assuming that "20" is a valid fan speed. In the long term, it would be useful to have a more robust solutions such as temperature averaging , smooth startup, conditional stopping of the fans or exposing the parameters to the user.

Linked is a GIF of the issue that I am trying to solve
fan_recording


jackpot51:
20% is too low for the 17 inch oryx. It's fans have a minimum speed of 25%

CourchesneA:
@jackpot51 Thanks for the info. I modified the PR to only change the temp where the fan starts from 55 to 50. The fan would still start at 25%, but it would solve the issue since I haven't seen my CPU go below 51 °C. The fan would always be on, but it seems a much better option than the constant toggling.

jackpot51:
I have an oryp7 and will try this out, thanks

jackpot51:
@curiousercreative we released #190 as of the 2021-07-20 release, which is available for most devices including the galp5.

CourchesneA:
In the end this pull request did not resolve the issue. It seems whatever the value I put in the fan curve, the fan will always reduce the temp of the CPU until it reaches the minimum value to trigger the fan, and the temperature will oscillate around that value, triggering the fan to go on and off continuously. I have tried multiple options including the `CFLAGS+=-DBOARD_HEATUP`, but the only way that I have found that was able to make the noise less annoying is by adding the fan point `FAN_POINT(35, 25)` to make the fan always active. One possible alternative that I can see would be raising the fan point high enough, but that is assuming that light operations do not raise the CPU temp enough to be dangerous or damageable. I do not know enough about CPUs to test this, so I will close the PR

PR #246, opened 2021-09-29
Fan Study

jackpot51:
Attempting to find and fix fan curve issues


leviport:
I've been running these fan curves for a while to get a feel for them. I'm still hearing the same faint little buzz noise when the fans start and stop, and this change doesn't seem to make it much less frequent. From all of the discussion in https://github.com/system76/ec/pull/190, it seemed like buzzing between 0% and 25% is basically inevitable (unless we can do some wizardy PWM stuff in that range). In that PR we decided that crossing the buzz range as quickly as possible was the best solution, so smoothing was disabled in that range. If buzzing is inevitable, we might have to prevent the fans from ever dropping below 25% if we want to get rid of it.

jackpot51:
@leviport I agree. The main issue is that no matter what, the spin up from stopped to any speed without buzzing will always buzz. We can try setting the minimum for the oryp* to 25%

PR #249, opened 2021-09-30
gaze16: set PCH_DPWROK_EC low on EC boot to ensure PCH is off

jackpot51:
This should make the gaze16 not require the AC adapter to be removed after flashing the EC.


crawfxrd (approved):
Would this potentially be a problem on the other TGL models?

jacobgkau (approved):
Confirmed this fixes the issue on gaze16-3060. Didn't see the issue happen on gaze16-3050 when flashing from the `flash.sh` script, but didn't see any regression there either.

jackpot51:
> Would this potentially be a problem on the other TGL models?

Yeah

PR #255, opened 2021-11-10
update deps to fix error in `make flash_internal`

curiousercreative:
I was not able to flash ec without re-generating the lockfile. In particular, building pkg-config was erring, so these lines may be all that is necessary: https://github.com/system76/ec/pull/255/files#diff-1e7cb82f2524f41246ba2496456a48e79f5323eb70b3850a1ae76be985e178f0R84-R86


crawfxrd:
Do you have what the error was?

tool/Cargo.lock
```diff
@@ -1,7 +1,5 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 3
```
crawfxrd, line 3: This shouldn't be getting removed: The new toolchain uses the v3 format.

curiousercreative: @crawfxrd how would I check that? I ran the deps script proof to this.

curiousercreative:
> Do you have what the error was?

It was the same error as https://github.com/rust-lang/pkg-config-rs/issues/124

tool/Cargo.lock
crawfxrd: `rustup self update`. Also check `cargo --version` to make sure the correct toolchain is being used (2021-06-12).

curiousercreative:
```bash
$ cargo --version
cargo 1.47.0-nightly (aa6872140 2020-07-23)

$ rustup self update
info: checking for self-updates
info: downloading self-update
  rustup updated - 1.24.3 (from 1.23.1)

$ cargo --version
info: syncing channel updates for 'nightly-2021-06-15-x86_64-unknown-linux-gnu'
info: latest update on 2021-06-15, rust version 1.55.0-nightly (539d7bd39 2021-06-14)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-src'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-src'
info: installing component 'rust-std'
info: installing component 'rustc'
info: installing component 'rustfmt'
cargo 1.54.0-nightly (44456677b 2021-06-12)
```

crawfxrd:
Error was a version of `rustup` that was too old to support the TOML toolchain file. As a result, the version of `cargo` and the Rust toolchain being used was not correct.

#267 will make is so running `deps.sh` will update rustup if it already available.

PR #261, opened 2021-12-15
battery: Group battery info in a struct

crawfxrd:


jackpot51 (dismissed):

jackpot51 (approved):

jacobgkau (approved):
This flashes on an oryp8, battery reporting & threshold configuration still work.

PR #263, opened 2021-12-21
Add FnLock key

crawfxrd:
Implement a FnLock toggle and a custom key for it.

- When disabled, `F1`-`F12` behave normally
- When enabled, `F1`-`F12` behave as the alternate function

Set it to `Fn+Esc` by default.

NOTE: Holding `Fn+Esc` at power on (S5 to S0) is used for resetting the EC configs. After S0, `Fn+Esc` will only act as FnLock.

NOTE: System76 laptops do not have an indicator for FnLock state.

Resolves: system76/firmware-open#78

### DIFF
--- a/src/board/system76/common/kbscan.c
+++ b/src/board/system76/common/kbscan.c
@@ -5,6 +5,7 @@
 #include <board/acpi.h>
 #include <board/fan.h>
 #include <board/gpio.h>
+#include <board/keymap.h>
 #include <board/kbc.h>
 #include <board/kbled.h>
 #include <board/kbscan.h>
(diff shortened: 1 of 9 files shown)


Stojek139808:
Hello, is there a reason why this is left as draft and not merged? I've tested this change on `clevo/ns70mu` and works correctly.

jackpot51 (dismissed):

leviport (approved):
This works great on darp8. 

I've opened https://github.com/pop-os/keyboard-configurator/issues/159 so we can track getting it added to the configurator.

leviport (approved):
Re-tested on darp8 and it's still working well. 

Sorry if my trying to update the branch messed something up :sweat_smile:

crawfxrd:
Default key has been changed from `Fn+Caps` to `Fn+Esc` to match what I see as the most common key used by other keyboards.

leviport (approved):
Working nicely on the lemp12. I confused myself after resetting the keyboard layout during testing, which made it stop working, but that's because the configurator doesn't yet support FnLock.

jackpot51:
I have concerns with the lack of an indicator. This is a great thing for users who want it, but for users who are unaware of this feature, accidentally toggling function lock could be very hard for them to undo.

crawfxrd:
I've dropped the commit adding it to the default layouts.

Users will have to explicitly add the key to their layout if desired.

leviport (approved):
Sounds good to me

jackpot51 (approved):

PR #265, opened 2022-01-04
arch/8051: Use Timer 2 for global tick

crawfxrd:
The ECs support Timer 2 from the 8052, which has a 16-bit auto-reload
mode. Use it to offload resetting the timer to hardware instead of doing
it in the interrupt.

### DIFF
--- a/src/arch/8051/time.c
+++ b/src/arch/8051/time.c
@@ -1,41 +1,41 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
-// Uses timer 0 to keep track of global time
-
-#include <8051.h>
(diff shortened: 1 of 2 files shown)


jackpot51:
Approved, but curious to know how you tested it

crawfxrd:
For now, just by running it.

The only difference should be removing a small amount of skew between intervals (a few dozen usec at most) as Timer 0 was stopped when reloading the values and updating `timer_overflows`.

jackpot51 (changes requested):
We will soon have an EC that only has the 8051 regs. I recommend either leaving the timing code as-is, or moving it to the ec/ite directory

PR #267, opened 2022-01-14
deps.sh: Update rustup or inform user of env vars

crawfxrd:
Ensure `rustup` is up-to-date to avoid users having a version that is too old to support the TOML format.

If rustup was just installed, tell the user to source the new env vars to use the correct version of the toolchain.

### DIFF
--- a/scripts/deps.sh
+++ b/scripts/deps.sh
@@ -23,7 +23,7 @@ if [[ "${ID}" =~ "debian" ]] || [[ "${ID_LIKE}" =~ "debian" ]]; then
         gcc \
         gcc-avr \
         libc-dev \
-		libhidapi-dev \
+        libhidapi-dev \
         libudev-dev \
         make \
         pkgconf \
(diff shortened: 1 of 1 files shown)


13r0ck:
This closes with the message `rustup was just installed. Ensure cargo is on the PATH with:` even if rust was never installed. I can still approve if you want, but it might be nice to also have the case that checks for rust not being installed?

crawfxrd:
If rustup is already available, it should print:

```bash
nightly-2021-06-15-x86_64-unknown-linux-gnu (overridden by '/data/Projects/System76/firmware-open/ec/rust-toolchain.toml')
rustc 1.55.0-nightly (539d7bd39 2021-06-14)

Successfully installed dependencies
```

If rustup is not available before running the script, it should print:

```bash
nightly-2021-06-15-x86_64-unknown-linux-gnu (overridden by '/data/Projects/System76/firmware-open/ec/rust-toolchain.toml')
rustc 1.55.0-nightly (539d7bd39 2021-06-14)

rustup was just installed. Ensure cargo is on the PATH with:
    source ~/.cargo/env

Successfully installed dependencies
```

13r0ck (approved):
Ah my bad. Yes, this is good :+1:

jackpot51 (approved):

PR #277, opened 2022-06-08
Add Darter Pro 8

crawfxrd:
system76/firmware-open#322

### DIFF
--- a/src/board/system76/darp8/board.mk
+++ b/src/board/system76/darp8/board.mk
@@ -1,0 +1,31 @@
+# SPDX-License-Identifier: GPL-3.0-only
+
+EC=it5570e
+
+# Enable eSPI
+CFLAGS+=-DEC_ESPI=1
+
+# Include keyboard
+KEYBOARD=15in_102
+
+# Set keyboard LED mechanism
+KBLED=rgb_pwm
+
+# Set battery I2C bus
+CFLAGS+=-DI2C_SMBUS=I2C_4
+
+# Set touchpad PS2 bus
+CFLAGS+=-DPS2_TOUCHPAD=PS2_3
+
+# Set smart charger parameters
+# Adapter input current = 4.74A
+# PRS1 = 0.010 ohm. Divide adapter input current by 2.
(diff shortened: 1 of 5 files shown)


src/board/system76/darp8/board.mk
```diff
@@ -0,0 +1,29 @@
...
+# Set touchpad PS2 bus
+CFLAGS+=-DPS2_TOUCHPAD=PS2_3
+
+# Set smart charger parameters
+#TODO: Find out why input current must by divided by two
+CHARGER=oz26786
+CFLAGS+=\
+	-DCHARGER_CHARGE_CURRENT=1536 \
+	-DCHARGER_CHARGE_VOLTAGE=8800 \
+	-DCHARGER_INPUT_CURRENT=2370
```
crawfxrd, line 26: Should we still be dividing by 2 for ADL?

jackpot51: Do you have a killawatt to test with?

jackpot51: Charging wattage (system off, battery drained) should be about 30W

jackpot51: Charging wattage should drop if the system is on and stress testing caused it to reach the max current to pull from the adapter.

jackpot51: System off, measure with killawatt. System on, measure total power with killawatt and charging power with sensors

crawfxrd: The killawatt is measuring ~16W with either value.

crawfxrd: Are we programming the oz26786 correctly?

The datasheet shows `AdapterCurrent()` uses each bit as a power-of-2 value to add to the current and not an immediate mA value.

crawfxrd: Okay, bit 0 is a 2mA weight. That's why it has to be divided by 2.

jackpot51 (approved):

jackpot51 (approved):

PR #285, opened 2022-06-17
Fix OZ26786 AdapterCurrent() values

crawfxrd:
The value passed to `AdapterCurrent()` must be adjusted based on the sense resistor used.

- 0.020 ohm sense resistor: Use actual input current
- 0.010 ohm sense resistor: Divide input current by 2
- 0.005 ohm sense resistor: Divide input current by 4

See section "Setting Adapter Current" in the datasheet.

Ref: OZ26786-DS v1.0

---

~I don't know if this is correct for gaze16, since it uses both a 5mΩ (PSR1) and 10mΩ (PRS2) sense resistor in the schematics, and we can never be sure that the schematics are correct.~

PRS1 is for `AdapterCurrent()` and PRS2 is for `ChargeCurrent()`, so this should be correct.

However, if the intention was to allow charging at 3A and not 1.5A, the value for `ChargeCurrent()` will need to be changed as well.

### DIFF
--- a/doc/adding-a-new-board.md
+++ b/doc/adding-a-new-board.md
@@ -2,33 +2,40 @@
 
 ## Charger parameters
 
-- `CHARGER_CHARGE_CURRENT`: Currently the same for all boards (1536).
+- `CHARGER_CHARGE_CURRENT`: Currently the same for all boards (3A).
 - `CHARGER_CHARGE_VOLTAGE`: On the battery, look for 充电限制电压 (charge limit
   voltage). Convert this from volts to millivolts.
 - `CHARGER_INPUT_CURRENT`: On the charger, look for DC output. Convert the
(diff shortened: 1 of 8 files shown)


crawfxrd:
The lemp9, lemp10, and darp7 have to divide for the same reason. They use a 5mΩ sense resistor instead of a 10mΩ like the other models.

crawfxrd:
Probably would be better to use the exact adapter input current and declare some RSENSE value so the charger code could handle the shift.

jackpot51 (approved):

crawfxrd:
For boards not using a 10mΩ adapter sense register, `ChargeOption` will need to be changed as well.

jackpot51 (approved):

crawfxrd:
Replaced by #288.

PR #289, opened 2022-07-07
gaze17-3050: Update GPIOs and configs

crawfxrd:
- system76/firmware-open#333

---

- Add to CI
- Correct smart charger values
- Enable WLAN power
- Enable USB power
- Disable mute
- Adjust fan points based on ODM-provided thermal data


XV-02 (approved):
Fan curves, USB power, etc. look to be working as desired. Approving.

13r0ck (approved):

PR #292, opened 2022-07-19
add lemp11 fan curve

n3m0-22:
This is a proposal to add a fan curve for the new Lemur Pro `lemp11`. I have been testing it for about a week and have seen significant improvements, approximately 20 degrees cooler, in temperatures both idling and under load. If this is too aggressive the values can always be changed, but I think it would be of value to add a curve that improves temperatures without adding much if any extra fan noise. So far I have not noticed a difference in terms of sound.


src/board/system76/lemp11/board.mk
```diff
@@ -28,5 +28,19 @@ CFLAGS+=\
...
+CFLAGS+=-DBOARD_COOLDOWN=20
+CFLAGS+=-DBOARD_FAN_POINTS="\
+	FAN_POINT(45, 25), \
+	FAN_POINT(50, 30), \
+	FAN_POINT(55, 30), \
+	FAN_POINT(60, 40), \
+	FAN_POINT(65, 50), \
+	FAN_POINT(70, 60), \
+	FAN_POINT(75, 70), \
+	FAN_POINT(80, 80) \
```
crawfxrd, line 42: These all increase duty by 10% for every 5°, so I think you only need the last one with  interpolation enabled.

crawfxrd:
> approximately 20 degrees cooler

Not surprising: the first point for the default curve is at 70°C, which is way too high. 

Some values from the provided thermal data for the proprietary firmware:

| Temp (°C) | Duty (%) | Noise (dB) |
| -- | -- | -- |
| 50 | 41 | 25 |
| 60 | 53 | 30 |
| 64 | 69 | 35 |
| 80 | 87 | 40 |

src/board/system76/lemp11/board.mk
n3m0-22: I've been testing this, and I get very similar temperature results with the following.

```
# Custom fan curve
CFLAGS+=-DBOARD_HEATUP=5
CFLAGS+=-DBOARD_COOLDOWN=20
CFLAGS+=-DBOARD_FAN_POINTS="\
	FAN_POINT(45, 25), \
	FAN_POINT(80, 80) \
"
```
I've been reading old PRs regarding interpolation, https://github.com/system76/ec/pull/190 for example, but I think I'm missing something. How do I enable interpolation?

When I try with just the last one I don't see nearly the same improvements in terms of temperature.

```
CFLAGS+=-DBOARD_HEATUP=5
CFLAGS+=-DBOARD_COOLDOWN=20
CFLAGS+=-DBOARD_FAN_POINTS="\
	FAN_POINT(80, 80) \
"
```

crawfxrd: Interpolation is enabled by default (`SMOOTH_FANS` is set to 1).

So

```
FAN_POINT(45, 25), \
FAN_POINT(50, 30), \
FAN_POINT(55, 30), \
FAN_POINT(60, 40), \
FAN_POINT(65, 50), \
FAN_POINT(70, 60), \
FAN_POINT(75, 70), \
FAN_POINT(80, 80), \
```

should be the same as

```
FAN_POINT(45, 25), \
FAN_POINT(50, 30), \
FAN_POINT(55, 30), \
FAN_POINT(80, 80), \
```

n3m0-22: Ok I guess I did understand it better than I thought. So if I only have the line `FAN_POINT(80, 80), \` the first point starts at the 70°C default, but if I add the `FAN_POINT(45, 25), \` it creates a new minimum which is why I'm getting the same results? If so would using just those two points be acceptable, or did I misunderstand and that's what you meant from the beginning?

crawfxrd: Setting these values overrides the defaults. So with _only_ `FAN_POINT(80, 80)`, the fan comes on at 80°C at 80% and stays at 80% until temps go below 80°C again. (There's the ramp-up/cool-down timings as well, but I'm ignoring those for now.)

n3m0-22: My reasoning for having the fan turn on at a lower temperature is due to the fact that it was idling around the mid 60's. I would not normally worry about a temperature this low, but to me it was getting hot to the touch on the upper left corner near the escape key and radiating out. This is on a small open stand that sets it at and angle of about 20-25 degrees. It runs hotter laying flat on a surface. This creates a noticeable difference in the temperature between the left and right side of the keyboard after a couple hours of use. Setting it to a lower minimum eliminates this, and keeps the idle temperature around the mid 40's. Is the concern with this that it will cause a shorter life span for the fans?

n3m0-22: Per your recommendation, I've been testing all day with only the single `FAN_POINT(80, 80)`. I have seen no change in the default behavior. The system still seems to run hot. It's only when adding the other value `FAN_POINT(45, 25)` that I see a temperature drop.

crawfxrd: That is not my recommendation.

I am saying that what you have is equivalent to:

```
FAN_POINT(45, 25), \
FAN_POINT(50, 30), \
FAN_POINT(55, 30), \
FAN_POINT(80, 80), \
```

n3m0-22: I'm sorry, that's my fault. I misunderstood. If I'm reading this correctly it should just be those values? If so I'll make that change right away. Did you want a change in the in the `HEATUP` and `COOLDOWN` values as well, or to remove them?

crawfxrd: Yes, you can just use those points.

Those values should be fine for `HEATUP`/`COOLDOWN`.

n3m0-22: Thank you for you help and patience on this. I appreciate it. Please let me know if any other changes are necessary.

curiousercreative:
I can't remember where it's documented, but as I understand it, we want laptops to not spin fans persistently (which will be the result of these fan curves, no?) when not subject to load. Ideally we [default fan curves](https://github.com/system76/ec/blob/master/src/board/system76/galp5/board.mk#L39-L43) to coincide with [system76-power power profile thermal targets](https://github.com/pop-os/system76-power/blob/master/src/daemon/profiles.rs#L309-L324) so that the battery saving profile is confidently silent. With my galp5, battery saving profile will throttle at 68C, just below where fans start spinning while balanced will throttle at 88C, right before a strong fan curve ramp.

n3m0-22:
If that is the case I can modify this because yes this does cause the fan to stay around 2400-2600 RPM. My issue would be that without the custom curve it seems to idle around the low to mid 60's, and with the first point for the default curve being at 70°C the custom curve will need to be somewhere below that to have any impact. That would mean the fan will be turning on and off all the time with any level of work. To me that would be annoying, but that's just my opinion.

curiousercreative:
@n3m0-22 is the problem with idling in the 60s that it burns skin or what?

WatchMkr:
The goal is to maximize performance while minimizing fan noise. Usually, for U class processors at least, fans shouldn't be necessary at idle. They're only necessary at idle if the thermal solution can't passively handle the thermal energy thus causing the temps to continuously rise. This shouldn't be the case on the Lemur.

n3m0-22:
I wouldn't say burns, but it was uncomfortable to use especially on the upper left side of the keyboard if your hand is resting there. I tried a stand a first to allow for more airflow but it did not make much difference. That's why I made this change to mine in the first place. Maybe this is unnecessary and just an issue for me. We can always hold off and see if users have the same complaint and if so have this ready to go.

n3m0-22:
It does stay in the mid 60s passively and does not keep rising. I just found it to be uncomfortable with all day use, but that may not be the case for most users in which case this can be disregarded.

curiousercreative:
@n3m0-22 wish I had a lemp11 to provide feedback with, but hopefully others will chime in. Regardless, I'd encourage you to work on a lemp11 system76-power profile to match your proposed fan curve. If it doesn't comfortably run fanless, so be it. Silence is the goal for the battery power profile, see how close you can get.

crawfxrd:
> The goal is to maximize performance while minimizing fan noise.

How minimal? We're provided thermal data for proprietary firmware, so we have some numbers already for fan duty vs noise. The fan tables usually start near a point that only produces 25 dB of noise (measured 25 cm from the unit IIRC).

> Usually, for U class processors at least, fans shouldn't be necessary at idle.

Even H class units can idle at 50°C (in hybrid graphics mode) if you let the fans cool to that.

But this is a very different requirement than:

> Silence is the goal for the battery power profile

hyperair:
Hello, just a lemp11 user chiming in here -- my lemp11 (i7-1255U) runs at 68C on the battery profile with light usage (just chatting on Slack), and while that's fine on a desk, it does burn the thighs, especially near the upper-centre part of the underside of the laptop (on the other side of the U key).

Here are some graphs of how the thermals are like on the `balanced` profile:

image

curiousercreative:
@hyperair I'm not a repo owner/admin, so I can't say for sure, but I'd guess this won't be prioritized. Modifying the fan curve in firmware and flashing isn't too difficult. You can also show some support for [this PR](https://github.com/system76/firmware-open/pull/365) that would make fan speed control possible in userspace without flashing firmware.

hyperair:
@curiousercreative Thanks for the PR link, that looks great!

> Modifying the fan curve in firmware and flashing isn't too difficult

Yeah I found [this guide](https://devimalplanet.com/guide-how-to-adjust-the-fan-curve-on-system76-laptops) and it looks fairly straightforward, I'm just a little worried that I might brick the laptop in the process.

curiousercreative:
@hyperair no guarantees of course, but I've flashed my firmware similar to that guide at least a hundred times. All of my typos lead to a build failure prior to flashing. I've never had any flashing errors.

hyperair:
@curiousercreative That's comforting to hear, thanks

PR #294, opened 2022-07-31
battery: Fix value to start/stop charging at

crawfxrd:
Change the condition, which is currently a level *at* which charging
will start/stop.

Per [sysfs](https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-class-power), these values are a level:

- *below* which charging will begin
- *above* which charging will stop

### Test

1. Set stop threshold
2. Drain battery ~5% below that value
3. Charge the battery

- Check: Battery shows the user specified value, not 1 below it

### DIFF
--- a/src/board/system76/common/battery.c
+++ b/src/board/system76/common/battery.c
@@ -58,7 +58,7 @@ int16_t battery_charger_configure(void) {
         // Stop threshold not configured: Always charge on AC.
         should_charge = true;
     }
-    else if (battery_info.charge >= battery_get_end_threshold()) {
+    else if (battery_info.charge > battery_get_end_threshold()) {
         // Stop threshold configured: Stop charging at threshold.
         should_charge = false;
     }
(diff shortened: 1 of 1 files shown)


13r0ck (dismissed):

linuxgnuru:
Set to 40 90, drained to 84 and charged up to 90 but after discharging again to 84 would not charge past 85 even though the charge thresholds were not changed

linuxgnuru (dismissed):
Tested 5 times with darp8, lemp11, and gaze17-3050; battery never got above or below the set thresholds

crawfxrd:
ThinkPads have the current behavior of stopping _as soon as_ it reaches the value. (Not surprising, since that's what I based the implementation on.)

What is the greater expectation:

- that this should behave just like ThinkPads
- or that this should behave as interpreted here from the kernel documentation?

(Additionally, I still don't know how any other vendor behaves. Does HP Dev One support charging thresholds? What does it do?)

jacobgkau:
> What is the greater expectation:
>     * that this should behave just like ThinkPads
>     * or that this should behave as interpreted here from the kernel documentation?

I think the expectation is that it behaves in a way that's easy for users to understand. I think it has come up a couple of times already that the percentage shown will often be one less than what's configured as the maximum because it stops charging immediately and then doesn't have enough charge to actually display the configured maximum percentage. I would expect this to come up more once we have a GUI to configure this and once it becomes persistent so it can be more useful. If a user configures a maximum value, they'll probably expect to see that maximum value being reached.

13r0ck (approved):

linuxgnuru (approved):

PR #295, opened 2022-08-02
board/system76/common: use SLP_S0# pin for modern standby detection

mkopec:
Previously, CPU_C10_GATE# was used for detecting entry into s0ix /
Modern Standby.

Intel documentation says that SLP_S0# should be used for s0ix detection
instead, and CPU_C10_GATE# is intended for VCCSTG power gating, so
switch the pin to improve s0ix detection reliability.

Fixes https://github.com/system76/firmware-open/issues/199

Tested on Clevo NS51MU, firmware built for `system76/darp7`.

References:

- Intel docs #575570 rev 0.5, #607872 rev 2.2
- Chromium EC: https://chromium.googlesource.com/chromiumos/platform/ec/+/master/power/intel_x86.c#41

Signed-off-by: Michał Kopeć <michal.kopec@3mdeb.com>

### DIFF
--- a/src/board/system76/common/peci.c
+++ b/src/board/system76/common/peci.c
@@ -123,8 +123,8 @@ uint8_t peci_get_fan_duty(void) {
     uint8_t duty;
 
 #if EC_ESPI
-    // Use PECI if CPU is not in C10 state
-    peci_on = gpio_get(&CPU_C10_GATE_N);
+    // Use PECI if platform is not in CS
+    peci_on = gpio_get(&SLP_S0_N);
 #else // EC_ESPI
     // Use PECI if in S0 state
     peci_on = power_state == POWER_STATE_S0;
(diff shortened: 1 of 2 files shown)


crawfxrd:
Verified behavior on lemp10.

Should be tested on:

- All TGL-U boards
- Any 1 TGL-H board
- lemp11
- Any 1 other ADL board

Test with/after https://github.com/system76/firmware-open/pull/341 for updated coreboot base.

mkopec:
Ah, I don't have access to these models' schematics so I didn't realize they didn't have `SLP_S0#` connected to the EC. Does it make sense to implement a function like `is_suspended` that does the right check for boards with `SLP_S0#`, `CPU_C10_GATE#` only, or with S3 enabled?

crawfxrd:
Can probably just define `SLP_S0_N` as the `CPU_C10_GATE_N` pin for those 2 in their `gpio.c` with a `FIXME`.

XV-02:
For some reason, this change on Lemp11 is causing a failure to post on only the integrated ram.

crawfxrd:
Merged with #298.

PR #296, opened 2022-08-02
galp6: Fix USB power

crawfxrd:
The schematics incorrectly show the pin as `USB_PWR_EN#`. There is a comment on the port side clarifying that it is actually active high.

Additionally, `PD_EN` is used instead of `VA_EC_EN` (like every other board) for enabling power to TCP0.


XV-02:
With or without this, devices which do not have independent power are not communicating with the left-side (charging) USB-C port. The same devices connected to a USB-C hub and then to the port work flawlessly.

XV-02 (approved):
All USB/Thunderbolt ports are now correctly functioning. Approving.

13r0ck (approved):

PR #297, opened 2022-08-03
galp6: Do not provide power while off

crawfxrd:
The previous commit incorrectly enabled power while off. Model `PD_EN` after `VA_EC_EN`, which it replaced on galp6 for TCP0.

Fixes: a6a6c5fba4fb ("galp6: Fix TCP0 power")

### Test

Power off and unplug the unit.

- Before change
  - Power LED blinks orange, indicating CPU is powered
  - Battery is drained over night
- After change
  - Power LED remains off
  - Battery is at the same capacity as before (probably <2% difference, due to powering off/on)

### DIFF
--- a/src/board/system76/common/power.c
+++ b/src/board/system76/common/power.c
@@ -73,6 +73,11 @@
     #define HAVE_VA_EC_EN 1
 #endif
 
+// Only galp6 has this, so disable by default.
+#ifndef HAVE_PD_EN
+    #define HAVE_PD_EN 0
+#endif
+
 #ifndef HAVE_XLP_OUT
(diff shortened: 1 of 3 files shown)


13r0ck (approved):

XV-02:
This seems to be resolving the powered EC issue. Not seeing the flashing light, nor am I seeing indications of the increased power draw that would be present with EC powered.

XV-02 (approved):
This seems to be resolving the powered EC issue. Not seeing the flashing light, nor am I seeing indications of the increased power draw that would be present with EC powered.

PR #298, opened 2022-08-04
Add config for using S0ix and fix detection

crawfxrd:
S0ix does not require eSPI, and eSPI does not mandate S0ix.

Enable S0ix for:

- darp7 (TGL-U)
- darp8 (ADL-P)
- galp5 (TGL-U)
- galp6 (ADL-P)
- lemp10 (TGL-U)
- lemp11 (ADL-U)

It is also currently enabled on ADL-H due to a bug with S3:

- gaze17-3050
- gaze17-3060-b
- oryp9

### DIFF
--- a/src/board/system76/common/peci.c
+++ b/src/board/system76/common/peci.c
@@ -127,8 +127,8 @@ uint8_t peci_get_fan_duty(void) {
     uint8_t duty;
 
 #if USE_S0IX
-    // Use PECI if CPU is not in C10 state
-    peci_on = gpio_get(&CPU_C10_GATE_N);
+    // Use PECI if platform is not in CS
+    peci_on = gpio_get(&SLP_S0_N);
 #else // USE_S0IX
     // Use PECI if in S0 state
     peci_on = power_state == POWER_STATE_S0;
--- a/src/board/system76/common/peci.c
+++ b/src/board/system76/common/peci.c
@@ -10,6 +10,10 @@
 #include <ec/gpio.h>
 #include <ec/pwm.h>
 
+#ifndef USE_S0IX
+    #define USE_S0IX 0
+#endif
+
 // Fan speed is the lowest requested over HEATUP seconds
 #ifndef BOARD_HEATUP
     #define BOARD_HEATUP 4
@@ -122,13 +126,13 @@ int16_t peci_wr_pkg_config(uint8_t index, uint16_t param, uint32_t data) {
 uint8_t peci_get_fan_duty(void) {
     uint8_t duty;
 
-#if EC_ESPI
+#if USE_S0IX
     // Use PECI if CPU is not in C10 state
     peci_on = gpio_get(&CPU_C10_GATE_N);
-#else // EC_ESPI
+#else // USE_S0IX
     // Use PECI if in S0 state
     peci_on = power_state == POWER_STATE_S0;
-#endif // EC_ESPI
+#endif // USE_S0IX
 
     if (peci_on) {
         // Wait for completion
--- a/src/board/system76/common/power.c
+++ b/src/board/system76/common/power.c
@@ -21,6 +21,10 @@
     #include <board/espi.h>
 #endif
 
+#ifndef USE_S0IX
+    #define USE_S0IX 0
+#endif
+
 #define GPIO_SET_DEBUG(G, V) { \
     DEBUG("%s = %s\n", #G, V ? "true" : "false"); \
     gpio_set(&G, V); \
@@ -313,7 +317,7 @@ static bool power_peci_limit(bool ac) {
 void power_set_limit(void) {
     static bool last_power_limit_ac = true;
     // We don't use power_state because the latency needs to be low
-#if EC_ESPI
+#if USE_S0IX
     if (gpio_get(&CPU_C10_GATE_N)) {
 #else
     if (gpio_get(&BUF_PLT_RST_N)) {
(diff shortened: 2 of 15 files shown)


13r0ck (approved):

XV-02 (dismissed):
On Lemp11, this is resulting in a failure to post without a non-soldered ram dimm.

crawfxrd:
That's unrelated. Most likely caused the coreboot rebase.

XV-02:
On Lemp11, I do not believe we're reaching s0ix.
Looking at Intel documentation, the files at `/sys/kernel/debug/pmc_core/slp_s0_residency_uses` and `/sys/devices/system/cpu/cpuidle/low_power_idle_system_residency_us` should have higher values after suspending to s0ix. However, neither seems to increase from `0` across suspend cycles.

Given the proximity of the current head for the coreboot submodule in firmware-open, I've been checking out coreboot from `system76` to ensure the presence of the mix-topo alderlake fix on Lemp11. I do not believe this should be causing issues, but want to mention it. Beyond that, however, I'm only checkout out the s0ix branch for EC when I'm building Lemp11.

crawfxrd:
lemp11 has the bug where the CPU doesn't reach C10 on battery power. You will need to test on AC power.

XV-02:
While some hardware compatibility tests are outstanding, and Lemp11 is experiencing a generalized issue regarding hitting the package C10 state during suspend, both Lemp 11 and Galp6 are currently looking good.

XV-02:
Just to formally finalize: 
Darp8, Galp6, and Lemp11 passed without regression.
Galp5 and Lemp10 are still awaiting testing.
Here are checklists for completed tests:
[Darp8-s0ix.md](https://github.com/system76/ec/files/9404222/Darp8-s0ix.md)
[Galp6-s0ix.md](https://github.com/system76/ec/files/9404223/Galp6-s0ix.md)
[Lemp11-s0ix.md](https://github.com/system76/ec/files/9404224/Lemp11-s0ix.md)

crawfxrd:
Are you testing with a USB-C device plugged in during suspend?

jacobgkau:
> Are you testing with a USB-C device plugged in during suspend?

No, that has never worked. Is it supposed to now?

crawfxrd:
Yes; it's the primary purpose of the change.

jacobgkau:
Just to be clear, this PR supersedes https://github.com/system76/ec/pull/295 and https://github.com/system76/ec/pull/295 will no longer need to be tested/merged after this one?

Suspending and resuming with USB-C devices plugged in does work on both darp7 and lemp10 now. Tested on each with a USB-C hard drive enclosure and a USB-C display.

However, both laptops fail to resume with a Thunderbolt hard drive enclosure connected. They enter suspend with the blinking LED, but upon trying to resume, the LED turns solid but the internal display never turns back on (switching to a TTY doesn't work in this state.) This failure only occurs when the drive is actually mounted; when it's not mounted, I can resume successfully, but I see/hear the OS disconnect and reconnect the drive afterwords.

The failure to resume with a mounted Thunderbolt drive is not a regression from master firmware; on the master branch, the system doesn't go into suspend, but still fails to wake up after attempting to suspend. However, it is a regression over currently released firmware-- currently released firmware doesn't suspend, but does eventually resume.

crawfxrd:
> Just to be clear, this PR supersedes #295 and #295 will no longer need to be tested/merged after this one?

Yes.

> However, both laptops fail to resume with a Thunderbolt hard drive enclosure connected.

Does this happen with a USB-C enclosure? What's the TBT enclosure?

jacobgkau:
> Does this happen with a USB-C enclosure?

No, the same enclosure connected via USB-C does not have the issue. It suspends/resumes successfully after the change (where it used to not suspend.)

> What's the TBT enclosure?

I think [this](https://www.amazon.com/G-Technology-G-DRIVE-Thunderbolt-Desktop-External/dp/B071RNNH52) might be the exact one. It's an aluminum G-Drive enclosure with one USB-C port and two Thunderbolt ports (I only ever connect one of these at a time.)

XV-02 (approved):
All the listed models seem to be suspending and resuming in an s0ix manner. Additionally, all the indicated models can suspend and resume with USB-C devices attached.

crawfxrd:
I only have the eGPU, but I do see it fail to go from D3 to D0. I think this will require changes to the FSP or coreboot config.

thomas-zimmerman:
I was testing this with a customer where we found that using the "Fn+1" fan toggle would cause a `lemp11` to fail to enter suspend state. The system looks like it suspends in the logs but the LED does not enter the suspend blink and checking the counter in `/sys/devices/system/cpu/cpuidle/low_power_idle_system_residency_us`. Closing the lid would sometimes allow the system to fully suspend and pressing the "Fn+1" key will turn off the fan in suspend.

Toggling the fan speed with "Fn+1" would also make it much more likely that the system would not reach suspend in future suspend attempts.

leviport:
Adding a little more context to Thomas's comment: he was testing this on the lab lemp11 which had this PR's EC flashed to it. It doesn't sound like this PR causes a regression, but the behavior sounded like it was worth reporting.

leviport:
@thomas-zimmerman just to clarify, are you saying that the behavior you mentioned is a regression in this EC update, or that it was present before this update, and still persists after this update?

thomas-zimmerman:
@leviport I was working with a customer that is on the current release firmware and I was testing on this PR's firmware that was on the lab machine. The testing that I was doing was happening on both machines.

leviport (approved):
Alright, we don't have to delay merging this if it isn't a regression.

PR #301, opened 2022-09-14
charger: Fix input current mask, removing using masks

crawfxrd:
Despite Kill-A-Watt reporting correct values while off, charging is apparently broken while the system is on. Do not mask values to restore previous behavior.

Also adjust Oryx Pro charge current, which was originally at 1.5A and changed to 3A, down to 2A. This makes them charge at ~31W.

Fixes: system76/firmware-open#356


leviport:
I flashed this EC on my darp5 last night, and charging was working as expected.

linuxgnuru:
Flashing on oryp8, oryp7, and gaze16-3060-b I've found it will show not charging if below 35% but will trickle charge and between 90% and 99% it will also show not charging but if the battery is at 89% it will be in the "charging" state until full-charge is reached. I will see if 90% up trickle charges.

linuxgnuru:
another oddity; if stress testing with s-tui/stress-ng when the battery is below 35% the system shuts off. I replicated this on the oryp8 and the gaze16-3060-b

crawfxrd:
Can you reproduce this behavior on 20d40f347741?

crawfxrd:
> it will show not charging if below 35% but will trickle charge

I can't reproduce this on oryp8.

I also *don't* have the problem on oryp6 (@leviport did), but I am currently using the oryp9 charger. I'm running it down now to test charging from a low capacity.

Looking at my PRs, I can't see the charging problems being anything else but the mask. I should have clamped the original value before masking it.

crawfxrd:
I see it stop charging/trickle charge at 99% on galp5 and oryp6. I think this happened before, and is some other function of the smart charger.

linuxgnuru:
Not sure if this is connected, but on the gaze16-3060-b over 4 days the charge was at 97% and didn't even trickle charge to 100% (i compared /sys/class/power_supply/BAT0/charge_now and the value did not change over the 4 days)

crawfxrd:
Replaced by #303.

PR #303, opened 2022-09-20
Fix charger values, again

crawfxrd:
- Limit charger values to max valid value
- bq24780s: Fix charge current mask
- bq24780s: Set RSENSE ratio option
- oryp[5-8]: Reduce charge current to 2A

Fixes: 6295f601720b ("Fix smart charger values")
Fixes: system76/firmware-open#356
Replaces: #301


crawfxrd:
That SDCC can't optimize `static const` kills me a little bit every time I work on the EC.

crawfxrd:
Charger still stops charging at 99%, but charging while on works.

linuxgnuru:
On the oryp7 90-99% is still not charging (not even trickle) and /sys/class/power_supply/BAT0/current_now is always 0

linuxgnuru:
oryp8 was charging if 90~98% initially but after recent commit it is back to only charging if it is under 89%

leviport:
John says the kernel definitely has something to do with it. He's seeing it work as expected on 5.18.10 and the current 6.0 rc in Mainline.

linuxgnuru:
Testing with kernel 6.0 RC3 also had no issues with the 90-100% however I am seeing the oryp8 consistently shutdown while trying to resume if the battery is under 25%

linuxgnuru:
Going through the changelog for 5.19 (https://mirrors.edge.kernel.org/pub/linux/kernel/v5.x/ChangeLog-5.19) I found the following:
```
commit 185d20694a8aceb4eda9fc1314cbaad0df0aab07
Author: Werner Sembach <wse@tuxedocomputers.com>
Date:   Wed Apr 27 17:40:53 2022 +0200

    ACPI: battery: Make "not-charging" the default on no charging or full info
    
    When the battery is neither charging or discharging and is not full,
    "not-charging" is a useful status description for the case in general.
    Currently this state is set as "unknown" by default, expect when this is
    explicitly replaced with "not-charging" on a per device or per vendor
    basis.
    
    A lot of devices have this state without a BIOS specification available
    explicitly describing it. e.g. some current Clevo barebones have a BIOS
    setting to stop charging at a user defined battery level.
    
    Signed-off-by: Werner Sembach <wse@tuxedocomputers.com>
    Signed-off-by: Rafael J. Wysocki <rafael.j.wysocki@intel.com>
    ```
    
Could this have any affect on our own implementation of charge thresholds?

linuxgnuru:
I am unsure what is different but today the other kernels are now showing the same as before so my theory of it being kernel related is false

jacobktm:
In just one instance, so far, the system shut down while unplugged, with the battery at 90% almost the instant I started running stress-ng. So far after restarting it has not done so again.

jacobktm:
I did some additional testing of this oryp8 and have found that it shuts down when I attempt to resume the system from suspend with the charger unplugged when the battery is up to 86% charge, above 86% charge the system resumes normally.

jacobktm:
I've installed Pop! OS 22.04 on this oryp8 and started testing. I charged the battery to 100% and now, while discharging, the battery percent is not dropping below 99%.

jacobktm:
The battery started discharging below 99% in Pop! and charges normally. Does not shut down when resuming from suspend. But it does shut down under load when the charger is not plugged in if the system is in the balanced power profile. It does not shut down under load in the battery life power profile.

jacobktm:
I reinstalled the current firmware and reinstalled Ubuntu and with the current firmware and Ubuntu 22.04 with kernel 5.19.0 the system does not shutdown when resuming from suspend. I'm going to test the current firmware with the 5.18.10 kernel and the firmware from this branch again with kernel 5.19.0 in Ubuntu next.

jacobktm:
I did a fresh Ubuntu 22.04 install and have re-tested with the firmware from this branch in the 5.18.10 kernel. The battery charges correctly. The system resumes from suspend while on battery power normally if the system was not under load before suspending. Otherwise, if the system was under load when suspending, it shuts down while resuming from suspend while on battery power. This behavior is consistent in Ubuntu 22.04 with kernel 5.19.0 and Pop! OS 22.04 with kernel 5.19.0. The system will also occasionally spontaneously shut down under load while on battery power or power off during the boot process while on battery power.

I'm going to test the firmware from this branch in conjunction with the [adl-pmc](https://github.com/system76/coreboot/pull/141) branch from coreboot

jacobktm (approved):
I've only tested this PR with oryp8, but this issue does fix the charging issues I was seeing. There is still an issue in which the system shuts down when resuming from suspend, while on battery power, if the system was under load when it suspended, and occasionally the system will still spontaneously shut down under load while on battery power. This shutdown behavior, however, is also present in the currently released firmware.

As this PR does fix the charging issue and does not create any new issues I think we should get this fix out ASAP.

XV-02:
I would also add that this improves charging behaviour on Oryp7.

13r0ck (approved):

crawfxrd:
Someone please disable squashing commits on this repo.

leviport:
:grimacing: sorry. I'll disable that now.

memjr:
Has there been any progress on this issue?  I have the latest firmware but I am not sure I understand what it was meant to do.  Seems the battery will not charge while plugged in unless it discharges to 89% or lower?

Is that how it is supposed to work?

13r0ck:
@memjr yes that is correct. If your charge threshhold is set to 90%.

A charge threshold protects your battery by only using a percentage of the battery, in this case 90%. So charging stops at 90%, anything above that would be passed the threshold, and would damage your battery at an increased rate

If you would prefer to charge to 100%, at the cost of the lifetime of your battery you can change that with the `system76-power` command

Issue #174, opened 2021-04-12
oryp6/7: Changing keyboard brightness also changes color when using custom colors


MilesBHuff:
*Cross-post from https://github.com/pop-os/system76-acpi-dkms/issues/8.*

**Distribution (run `cat /etc/os-release`):**
```
NAME="Manjaro Linux"
ID=manjaro
ID_LIKE=arch
BUILD_ID=rolling
PRETTY_NAME="Manjaro Linux"
ANSI_COLOR="32;1;24;144;200"
HOME_URL="https://manjaro.org/"
DOCUMENTATION_URL="https://wiki.manjaro.org/"
SUPPORT_URL="https://manjaro.org/"
BUG_REPORT_URL="https://bugs.manjaro.org/"
LOGO=manjarolinux
```
 
 
**Related Application and/or Package Version (run `apt policy $PACKAGE NAME`):**
`system76-acpi-dkms 1.0.2+1+g0bc966c-1`

**Issue/Bug Description:**
Setting the keyboard of the oryp7 to anything other than `ffffff` results in completely random colors being assigned.

**Steps to reproduce (if you know):**
Set keyboard brightness to anything less than 100% (I use 20%):
```
sudo su -c 'echo 667788 > /sys/class/leds/system76_acpi::kbd_backlight/color'
# Repeat the above several times.
```

**Expected behavior:**
The setting exposed by `system76-acpi-dkms` should use the assigned hex color.

**Other Notes:**
From @jacobgkau:
> It seems like full brightness is reliable. Maybe this bug has something to do with how the black color is mixed in to lower the brightness. Certain colors (such as plain red) do work, others like `310086` purple are only broken on specific brightness levels.

From @leviport:
> I am seeing the same behavior on oryp6 when I set the backlight color to `667788` and change the brightness with the brightness hotkeys. Here is a short video of the color changing to all kinds of different colors when repeatedly moving the brightness up and down by one keypress: https://user-images.githubusercontent.com/13512097/114424105-0b6b6a80-9b75-11eb-950e-04a853f8a7df.mp4

MilesBHuff:
Any updates on this?  Keyboard brightness is completely broken, lol.  Defeats the purpose of having brightness control buttons on the numpad.

jacobgkau:
> Keyboard brightness is completely broken, lol.

[...]

PR #304, opened 2022-10-11
pwm: Reload counters when they reach 0

crawfxrd:
Add IT5570E register that controls when the down counters are updated. Set them to update when they reach 0, instead of immediately when DCRi is written.

Fixes keyboard color changing when changing brightness levels if not using 0xFF for an RGB value.

Fixes: #174

### DIFF
--- a/src/board/system76/common/pwm.c
+++ b/src/board/system76/common/pwm.c
@@ -26,6 +26,11 @@ void pwm_init(void) {
     // Turn off CPU fan (temperature control in peci_get_fan_duty)
     DCR2 = 0;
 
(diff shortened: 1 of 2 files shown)


jacobgkau (approved):
Recreated https://github.com/system76/ec/issues/174 on oryp8 and confirmed this PR fixes it. (When using `310086` purple as an example, I do still see a slight shift in hue as I turn the brightness all the way from one side to the other; but the shift is now consistent, rather than jumping around, and it remains somewhat purple rather than jumping between blue and pink.)

13r0ck (approved):

PR #305, opened 2022-10-12
HACK: pmc: Add hack to force S0ix entry on lemp11

crawfxrd:
lemp11 sometimes fails to reach C10 and gets stuck in C0 during s2idle. For some reason, sending a PMC SCI allows the CPU to go to C10.

Requires: system76/coreboot#142


jacobktm (approved):
This works in Linux. It still doesn't work in Windows, but this behavior is consistent with the officially released firmware.

jackpot51 (approved):

PR #306, opened 2022-10-13
tool: Update dependencies

crawfxrd:
clap is only updated to 3.2 instead of 4.0 because I didn't want to deal with all the changes.

### DIFF
--- a/tool/Cargo.lock
+++ b/tool/Cargo.lock
@@ -3,15 +3,6 @@
 version = 3
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
-dependencies = [
- "winapi",
-]
-
-[[package]]
 name = "atty"
 version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


jackpot51 (dismissed):

jacobgkau (changes requested):
Tested on oryp9. Building still works, but after this change, flashing with `./scripts/flash oryp9` from the firmware-open directory fails at the EC step with this error:

```
SPI VERIFY: 32 MB
  Downloaded hidapi v1.4.2
  Downloaded os_str_bytes v6.3.0
error: failed to parse manifest at `/home/system76/.cargo/registry/src/github.com-1ecc6299db9ec823/os_str_bytes-6.3.0/Cargo.toml`

Caused by:
  feature `edition2021` is required

  consider adding `cargo-features = ["edition2021"]` to the manifest
```

crawfxrd:
The toolchain file in firmware-open is updated by https://github.com/system76/firmware-open/pull/304.

Alternatively, `flash.sh` should be changed to use `pushd`/`popd` so the EC project toolchain is used.

jacobgkau:
> The toolchain file in firmware-open is updated by [system76/firmware-open#304](https://github.com/system76/firmware-open/pull/304).

I see. That PR's marked as draft; I can take a look at this again either when it's marked ready for review or when we have a PR to make the `flash.sh` change you mentioned as an alternative.

jackpot51 (approved):

jackpot51 (dismissed):

jackpot51 (dismissed):

jackpot51 (approved):

jacobgkau (approved):
Builds & flashes now (tested on darp7).

PR #311, opened 2022-10-26
Replace makebin with objcopy

crawfxrd:
Replace custom SDCC tool with the more standard binutils tool.

`scratch.rom` and `flash.rom` are different (no unnecessary padding), but resulting `ec.rom` is the same.

### DIFF
--- a/src/arch/8051/arch.mk
+++ b/src/arch/8051/arch.mk
@@ -14,7 +14,7 @@ sim: $(BUILD)/ec.rom
 # Convert from Intel Hex file to binary file
 $(BUILD)/ec.rom: $(BUILD)/ec.ihx
 	@mkdir -p $(@D)
-	makebin -s $(CODE_SIZE) -p < $< > $@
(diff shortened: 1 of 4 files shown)


src/arch/8051/arch.mk
```diff
@@ -14,7 +14,7 @@ sim: $(BUILD)/ec.rom
 # Convert from Intel Hex file to binary file
 $(BUILD)/ec.rom: $(BUILD)/ec.ihx
 	@mkdir -p $(@D)
-	makebin -s $(CODE_SIZE) -p < $< > $@
```
crawfxrd, line 17: `-s $(CODE_SIZE)` did absolutely nothing because it was immediately followed by `-p`.

If we want to pad the ROM, `--pad-to=$(CODE_SIZE)` can be used. I would recommend padding to the full ROM size though (128k for ITE, 160k(?) for ENE).

jackpot51 (approved):

jacobgkau (approved):
Still able to build & flash (tested on oryp9).

PR #312, opened 2022-11-01
Add other launch keyboards

13r0ck:
Adds newer launch versions to ec tool.

`sudo ./system76_ectool --access hid info` used to only work launch_1. This should now work with all current lanuch products.

Can be tested by plugging in each launch and running the above command, and/or
```bash
cargo build --release
sudo ./target/debug/system76_ectool --access hid led_mode 0 10 110
sudo ./target/debug/system76_ectool --access hid led_save
```
then plug and replug keyboard. LEDs should have saved.

### DIFF
--- a/tool/src/main.rs
+++ b/tool/src/main.rs
@@ -402,7 +402,13 @@ fn main() {
                         #[allow(clippy::single_match)]
                         match (info.vendor_id(), info.product_id(), info.interface_number()) {
                             // System76 launch_1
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

leviport (approved):
Tested on launch_1, launch_2, launch_lite_1, and launch_heavy_1. I'm able to set and save LEDs on all four models.

PR #316, opened 2022-11-17
Set wireless power at init and power off

crawfxrd:
Add a new wireless object for controlling WLAN power state. Change the power sequence to enable WLAN at board init and disable it at power off.

Newer galp5 units sold to customers have an issue where they do not fully power off. This is apparently somehow caused by `WLAN_PWR_EN`.

The unit received for development do not have this issue.

### Test

- Ensure WiFi and Bluetooth still work
- Ensure power LED doesn't blink after power off on battery power

### DIFF
--- a/src/board/system76/addw1/board.c
+++ b/src/board/system76/addw1/board.c
@@ -4,6 +4,7 @@
 #include <board/gpio.h>
 #include <board/kbc.h>
 #include <board/power.h>
+#include <board/wireless.h>
 #include <common/debug.h>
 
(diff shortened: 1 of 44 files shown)


crawfxrd:
I opted for adding a `wireless` object instead of cramming even more stuff into `power`.

leviport (approved):
Confirmed to fix the problematic galp5 I have here in the lab!

jackpot51 (approved):

PR #317, opened 2022-11-17
Conditionally compile eSPI support

crawfxrd:
A board may use either the LPC bus or the eSPI bus. Only include eSPI support for boards that use it.


jackpot51 (approved):

src/board/system76/darp8/board.mk
```diff
@@ -7,12 +7,13 @@ EC=ite
 CONFIG_EC_ITE_IT5570E=y
 
 # Enable eSPI
-CFLAGS+=-DEC_ESPI=1
-# Apply PMC hack for S0ix
-CFLAGS+=-DPMC_S0IX_HACK=1
+CONFIG_BUS_ESPI=y
+CFLAGS+=-DCONFIG_BUS_ESPI=1
```
jackpot51, line 11: This CFLAG addition could be added in common.mk when CONFIG_BUS_ESPI=y is set

crawfxrd: Done

jackpot51 (approved):

PR #318, opened 2022-11-18
Enable WLAN at power_on() instead of board_init()

crawfxrd:
If the board is on AC power when powered off the EC will not reset, and WLAN power will not be enabled on next boot. Move enabling WLAN from `board_init()` to `power_on()`

Fixes: be4659a0cb6f ("Set wireless power at init and power off")


leviport:
Tested the problematic galp5 we have in the lab. It powers off correctly on battery power and has wireless after booted on AC.

jacobktm (approved):
I tested this on lemp11, gaze17-3050, and another of our galp5's and this fixes the issue with wifi on all three of those.

13r0ck (approved):

PR #319, opened 2022-11-21
Add INTC and WUC registers

crawfxrd:
Definitions for the INTC and WUC modules.

- IT8587E has INTC groups 0-18
- IT5570E has INTC groups 0-21
- WUC groups 1, 3, 4 have an extra register for enabling interrupts to INTC

### DIFF
--- a/src/ec/ite/include/ec/intc.h
+++ b/src/ec/ite/include/ec/intc.h
@@ -1,0 +1,137 @@
+// SPDX-License-Identifier: GPL-3.0-only
+
+// Interrupt Controller (INTC)
+
+// ISR: Interrupt Status Register
+// IER: Interrupt Enable Register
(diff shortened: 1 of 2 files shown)


XV-02:
As this PR only contains new files which, as far as I can tell, are yet to be references elsewhere; and all it does is expose the various EC interrupts, without using them yet; I should be able to make sure that the this EC builds and flashes without issue as a valid test that everything looks good from a QA perspective. Is that a reasonable read?

crawfxrd:
I'd say there's nothing to test. They're not included anywhere so the binaries are identical.

```
git checkout master
make BOARD=<b> VERISON=1
mv build/<b>/1 build/<b>/0
git checkout intc-reg
make BOARD=<b> VERISON=1
diff -r build/<b>/0 build/<b>/1
```

XV-02 (approved):

PR #320, opened 2022-11-21
power: Update LEDs in separate functions

crawfxrd:
Split some of the logic in `power_event()` to separate functions.

### DIFF
--- a/src/board/system76/common/power.c
+++ b/src/board/system76/common/power.c
@@ -360,6 +360,84 @@ static bool power_button_disabled(void) {
     return !gpio_get(&LID_SW_N) && gpio_get(&ACIN_N);
 }
 
+/**
+ * Update power LEDs.
+ */
+static void power_update_power_leds(bool ac_new) {
+    static uint32_t last_time = 0;
+    uint32_t time = time_get();
+
+    if (power_state == POWER_STATE_S0) {
+#if CONFIG_BUS_ESPI
+        // HOST_C10 virtual wire is high when CPU is in C10 sleep state
+        if (vw_get(&VW_HOST_C10) == VWS_HIGH) {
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

XV-02:
Using a Galp6 as a test system, I noticed that holding the power-button until the machine powered off resulted in the power-indicator light flashing amber (both on AC and on battery). This is not the same behaviour as if a shutdown command is passed to the system. When shutdown by passing `shutdown now`, the power light turned off as expected. 

I also noticed that short-stroking the power button after shutting down from the OS triggered the power LED to flash amber, though it did not trigger the boot sequence. Similar button presses on currently released firmware had no effect. 

Plugging in the power also caused the power indicator LED to start flashing amber with this branch. 

Also, when plugging in AC while the system was shutdown, the battery indicator lingered on green for 1-2 seconds before correctly shifting to amber to indicate charging.

XV-02:
Having revisited this since the latest round of commits, all the issues previously mentioned are still present.

crawfxrd:
```diff
--- build/system76/darp8/0/ec.mem	2023-07-24 12:57:35.640394578 -0600
+++ build/system76/darp8/1/ec.mem	2023-07-24 12:58:04.484213031 -0600
@@ -25,5 +25,5 @@
    Name             Start    End      Size     Max
    ---------------- -------- -------- -------- --------
    PAGED EXT. RAM                         0      256
-   EXTERNAL RAM     0x0001   0x050c    1292     2048
-   ROM/EPROM/FLASH  0x0000   0x8b21   35085    65536
+   EXTERNAL RAM     0x0001   0x050e    1294     2048
+   ROM/EPROM/FLASH  0x0000   0x8af0   35036    65536
```

PR #327, opened 2023-01-11
make: Enable compiling asm files

crawfxrd:
Allow specifying asm files the same way C files are added:

    board-common-y += foo.asm

The file *must* use an `.asm` extension for 8051. The toolchain file filters the sources to compile them into objects correctly, as the C compiler doesn't handle asm files itself.


jackpot51 (approved):

jackpot51 (approved):

jacobgkau (approved):
Building & flashing are unaffected (tested on oryp8).

PR #328, opened 2023-01-11
avr: Fix compiling with GCC 12

crawfxrd:
Compiling the Arduino targets with GCC 12 fails with the following:

    error: array subscript 0 is outside array bounds of 'volatile uint8_t[0]' {aka 'volatile unsigned char[]'} [-Werror=array-bounds]

Apply the workaround from the bug discussion to fix it.

Ref: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105523

### DIFF
--- a/src/arch/avr/i2c.c
+++ b/src/arch/avr/i2c.c
@@ -40,7 +40,7 @@ int16_t i2c_start(struct I2C *i2c, uint8_t addr, bool read) {
         return -1;
 
     // check if the device has acknowledged the READ / WRITE mode
-    uint8_t twst = TW_STATUS & 0xF8;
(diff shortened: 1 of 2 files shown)


jackpot51 (approved):

jacobgkau (approved):
I was unable to recreate the build error using GCC and g++ 12 on Pop!_OS, using arduino/mega2560 as a test target. However, the change did not cause any problems building.

ilikenwf:
@crawfxrd sorry for all the issues!

In this case, this should probably also apply to gcc 13 or gcc >=12 as I had to also use this CFLAG when compiling with v13.

PR #324, opened 2023-01-09
Update lints and run at pre-commit

crawfxrd:
Rewrite the lint scripts. They are now named `\d\d-<name>.sh`. They will print what they do and whether the check passes. In the case they fail list the files that caused the failure, except for clang-format, which will be slow to run of every file individually (should just run `make fmt` anyway).

- `01-spdx-tags.sh`: Check all files have SPDX tags (except those explicitly excluded)
- `02-clang-format.sh`: Check formatting for all C source files and headers
- `03-shellcheck.sh`: Check shell files
- `lint.sh`: Run all lints

Lints run quickly, minimizing impact during development:

```
$ time ./scripts/lint/lint.sh 
Checking for SPDX tags...PASSED
Checking C style...PASSED
Checking shell scripts...PASSED

real	0m1.179s
user	0m0.762s
sys	0m0.496s
```

Problems prevent committing, ensuring they are fixed up front, not later:

```
$ git commit -m test
Checking for SPDX tags...PASSED
Checking C style...FAILED
Checking shell scripts...PASSED

issues found, not committing
```

Adds the new make targets:

- `list-boards`: List available build targets
- `fmt`: Apply clang-format
- `help`: Short description of make targets
- `git-config`: Sets up git hooks

PR #329, opened 2023-01-13
Generate and use compiler dep rules

crawfxrd:
SDCC supports the `-MMD` option. Use it instead of having every file depend on *all* the headers. Reduces incremental build time after modifying headers as it will now only rebuild the actual dependents and not the entire project.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,4 +96,4 @@ jobs:
         run: ./scripts/deps.sh
 
       - name: Build firmware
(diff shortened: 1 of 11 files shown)


XV-02 (approved):
I'm not seeing any issues with this. Approved.

jackpot51 (dismissed):

crawfxrd:
> Make-quiet needs testing and merging first

The base will update to `master` if this is merged.

There is also a conflict with #324, so this will require rebasing anyway.

jackpot51 (approved):

XV-02 (approved):
Looks good.

PR #332, opened 2023-01-18
Move gpio_debug from board to EC code

crawfxrd:
The `gpio_debug()` functionality depends on the ITE registers and not anything board-specific.


jackpot51 (approved):

XV-02 (approved):
Everything looks to build/work as desired. Approved.

PR #333, opened 2023-01-19
Remove PMC hack for S0ix

crawfxrd:
Whatever the issue was appears to be fixed after rebasing coreboot on the 4.19 release. lemp11 has successfully reached S0ix with `SLP_S0#` asserted when booted on battery power for 100 cycles.

Requires: https://github.com/system76/coreboot/pull/160

### DIFF
--- a/src/board/system76/common/acpi.c
+++ b/src/board/system76/common/acpi.c
@@ -19,8 +19,6 @@ extern uint8_t sci_extra;
 
 enum EcOs acpi_ecos = EC_OS_NONE;
 
-extern bool pmc_s0_hack;
-
 static uint8_t fcmd = 0;
 static uint8_t fdat = 0;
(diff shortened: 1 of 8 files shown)


jackpot51 (approved):

leviport:
I flashed the latest `master` of firmware-open and then flashed this EC afterwards on my oryp9. Unfortunately it looks like S0ix still isn't working on this unit. The machine acts like it's going to suspend (keyboard backlight turns off, and screens lock and turn off), but never fully gets there. The fans stay running, the power button light stays on, and the front indicator light remains solid. Also, `/sys/kernel/debug/pmc_core/slp_s0_residency_usec` still says `0` when I bring it back and unlock the screen to check that file.

leviport (approved):
Seems the oryp9 suspend bug is not related to this PR. Other than that, this seems to be working well.

PR #337, opened 2023-02-14
Update Rust toolchain to 2023-01-21

crawfxrd:
Update the toolchain and deps, and use the sparse registry access.


jackpot51 (approved):

leviport (approved):
I successfully built a half-dozen or so arbitrary models successfully on my main machine, and I was also able to build and flash on a gaze15. Everything is building and flashing nicely.

PR #338, opened 2023-02-17
peci: Fix waiting for completion and error checking

crawfxrd:
- Check `FINISH` instead of `HOBY` for command completion
- Check hardware error bits (2, 3, 5, 6, 7) for hardware error
- Clear the status after command completion
- Log `WrPkgConfig()` errors in peci instead of power

TODO:

- Check if `peci_wr_pkg_config` error still happens when booting on battery power

### DIFF
--- a/src/board/system76/common/peci.c
+++ b/src/board/system76/common/peci.c
@@ -295,8 +295,8 @@ bool peci_get_temp(int16_t *data) {
     // Start transaction
     HOCTLR |= 1;
 
-    // Wait for completion
-    while (HOSTAR & 1) {}
+    // Wait for command completion
+    while (!(HOSTAR & BIT(1))) {}
 
     uint8_t status = HOSTAR;
     if (status & 0xEC) {
@@ -346,8 +346,8 @@ int16_t peci_wr_pkg_config(uint8_t index, uint16_t param, uint32_t data) {
     // Start transaction
     HOCTLR |= 1;
 
-    // Wait for completion
-    while (HOSTAR & 1) {}
+    // Wait for command completion
+    while (!(HOSTAR & BIT(1))) {}
 
     uint8_t status = HOSTAR;
     if (status & 0xEC) {
--- a/src/board/system76/common/peci.c
+++ b/src/board/system76/common/peci.c
@@ -301,12 +301,17 @@ bool peci_get_temp(int16_t *data) {
     uint8_t status = HOSTAR;
     if (status & 0xEC) {
         ERROR("peci_get_temp: hardware error: 0x%02X\n", status);
+        // Clear status
+        HOSTAR = HOSTAR;
         return false;
     } else {
         // Read two byte temperature data if finished successfully
         uint8_t low = HORDDR;
         uint8_t high = HORDDR;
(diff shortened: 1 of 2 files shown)


jackpot51 (dismissed):

XV-02:
How should I test this? Build, flash, and boot to check for regressions; or am I looking for a specific output somewhere?
Looking at the PECI kernel docs doesn't suggest any sysfs exposure. Does this need serial debugging?

crawfxrd:
Check for regressions related to CPU fan behavior; including under load and after suspend (both S3 and Modern Standby).

XV-02:
I want to double check it wasn't a situation unique bad-flash, but attempting to build Lemp9 firmware with this present resulted in a system which doesn't seem to be booting, and certainly is not booting successfully. 

On Galp6, CPU fan behaviour looked good, but I did notice that prior to this, Galp6 had a tendency to blast the fans when first spinning them up, then to reel them back in. With this, the fans come on at a lower RPM, but seem to keep the system at a comparable temp. I'm assuming this is desired given conversations elsewhere?

crawfxrd:
Is lemp9 broken with this change specifically, or on `master`?

>  I'm assuming this is desired given conversations elsewhere?

What conversations?

Fan behavior may be different since it now waits for the host to complete the command instead of only a transaction, but there should be no significant change.

XV-02:
> What conversations?

Sorry, thought I'd seen mention somewhere of changing fan behaviour away from hitting full speed when first activated. 

Thermal performance seems effectively consistent, but on currently released Galp6 firmware, the fans throttle to max when they start, and then reign back in. Building from open-firmware master with this branch checked out in EC, they start and throttle to meet the thermal need. 

On Lemp9, the failure was with firmware-open master with this branch checked out in EC. Right now I'm checking against a clean firmware-open build, and if that operates as desired will check the firmware-open + PECI build again. 

Once I've tested that, I want to check more performant system using both s3 and modern standby. Seems that IDing whether Lemp9 is generally having a problem with this is probably the best next step though.

XV-02:
The Lemp9 problem is not directly tied to this PR, and I've opened a [dedicated issue](https://github.com/system76/firmware-open/issues/393) for the problem in firmware-open. 

I'll continue to look at this on a newer system utilizing s3.

XV-02:
On Oryp8, I was seeing an absolute cavalcade of tpm related errors while running a 150-cycle suspend test.
All of the errors were the following:
```
 A TPM error (6) occurred attempting get random
```
I've yet to be able to trip those errors with shorter (25 and 50 cycle) suspend tests. 
Mostly, it was just a test I ran on the off at the end of last night. I don't think this PECI change is the cause (can't see anything in the file changes that feels like it should impact it)  but, wanted to mention it in case I'm missing something.

Beyond that, Oryp8 looked like its fan behaviour was as expected. So far, other systems I've built this for also are behaving. But, I would ideally like to be able to test this on 10th gen systems before we merge it into master. But, beyond that, I would approve merger if _needed_.

jackpot51 (approved):

jackpot51:
We are merging this into the `rpl` branch

PR #340, opened 2023-03-10
Set PL4 on CPU reset

crawfxrd:


jackpot51 (approved):

leviport (approved):
This fixes battery overdraw after suspend/resume for me, as well as battery overdraw and EC lockup after the Plymouth decrypt prompt while booting on battery.. I got stuck in S3 for the first suspend after flashing firmware and EC, but that's a separate issue.

PR #344, opened 2023-03-27
power: Do not reset KBLED state on CPU wake

crawfxrd:
Do not reset the keyboard backlight on CPU wake. This fixes keyboard backlight turning off and becoming white on wake from sleep.

TODO: Fix for all kbled mechanisms.

### DIFF
--- a/src/board/system76/common/power/intel.c
+++ b/src/board/system76/common/power/intel.c
@@ -11,7 +11,6 @@
 #include <board/fan.h>
 #include <board/gpio.h>
 #include <board/kbc.h>
-#include <board/kbled.h>
 #include <board/lid.h>
 #include <board/peci.h>
 #include <board/pmc.h>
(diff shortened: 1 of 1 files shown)


crawfxrd:
Arguably, a function called "reset" *should* put the backlight level and color in a known state.

But what do bonw14, darp5, and oryp5 *actually* need to do for I2C when waking the CPU? Does `kbled_reset` need to be called at all?

ilikenwf:
In my tinkering this appeares to break the system76-acpi-dkms module, making the LEDs unusable with the hardware controls inside Linux. If you remove the module they begin to work again.

ilikenwf:
I may have spoken too soon... if I was wrong, I'll post about it again here...but seems my distro is not using the master branch.

ilikenwf:
So it does not appear to be this commit, however the current master does seem to break things with the keyboard backlight controls when the acpi module is in use.

PR #345, opened 2023-03-29
ci: Generate board matrix from directory contents

Yethal:
Use jq to generate json containing board/vendor combinations, then generate board matrix from it.
Combinations excluded from the matrix are stored in scripts/boards.sh


crawfxrd:
Can the existing make target `list-boards` be used?

Yethal:
Done

crawfxrd (approved):
Can we go a week without GitHub Actions shitting itself?

Anyways, I'm not an admin so I can't view or configure any project settings.

Issue #331, opened 2023-01-14
Oryx8 - Laptop suddenly dies if stressed on battery


rnd-ash:
Updated to latest EC and Coreboot version via system76-firmware-cli , and even tried the latest EC git code and have the same results.

Side note. I am using optimus-manager here to disable the DGPU on battery, will post an update comment if enabling the DGPU on battery fixes this to narrow down the issue.

When on battery power after a fresh boot, the laptop works fine, I can compile code and stress the system without issue. But simply after sleeping the notebook then resuming it, the laptop suddenly dies if anything moderately stressful is executed (Even launching chrome causes it!)

Once dead, the power button does nothing for around 20-30 seconds, then it responds and lets the notebook power up again.

rnd-ash:
I enabled the DGPU on battery (Switched from optimus-manager to system76-power and used hybrid graphics mode), and get the same result. If anything the laptop might be a bit more sensitive after sleeping (Lighter actions cause it to die), so maybe its a power budget issue after the system has resumed from sleep?

rnd-ash:
Another update. If I resume the notebook with the AC adapter plugged in, then unplug the adapter once the notebook has woken up, the issue does not occur when stressing it.

crawfxrd:
Able to reproduce on firmware 2022-11-21_b337ac6, which is what everything other model is using for the release.

crawfxrd:
Introduced by #298.

rnd-ash:
> Able to reproduce on:
> 
> * 2022-11-21_b337ac6, which is what everything other model is using for the release
> * EC `master` with coreboot 4.19 rebase
> 
> Does not happen with:
> 
> * coreboot 4.19 and [d3ee037](https://github.com/system76/ec/commit/d3ee0378792c4afbe069d6d8ed8c02f2eef2a0cc) (commit used for the 2021-09-23_b9b0e89 release)
> * coreboot 4.19 and [20d40f3](https://github.com/system76/ec/commit/20d40f347741880ee27757bf368691b82cc60bfa)
> 
> Based on the sound the fans make, I think the power limits are lost after S3. With [d3ee037](https://github.com/system76/ec/commit/d3ee0378792c4afbe069d6d8ed8c02f2eef2a0cc) they don't spin up while building the kernel (implying the 28W limit is still set), whereas on `master` they spin up pretty soon in the build process.

Can confirm. With i7z open, even with governer set to powersave, I see the CPU jump to 4100Mhz just before the notebook dies.

[...]

PR #348, opened 2023-05-01
Increase Oryp8 CPU Battery power limit to 45W

rnd-ash:
This massively improves system responsiveness (CPU can now run at 2.1Ghz constantly rather than locked at 1.5Ghz when on battery), does not put the laptop in a current overdraw situation resulting in the system powering off like mentioned in #331


jackpot51:
@crawfxrd @system76/quality-assurance I believe we can accept this change. We have recently raised the DC PL4 on RPL models to 55W. This change would be in-line with that for prior models.

crawfxrd (approved):

crawfxrd:
Done in #353.

PR #351, opened 2023-05-10
german keyboard layout + mux key kombo + smoother keyboard backlight + more status led options

Vitamin-X:
i hope i didnt missed any


crawfxrd:
There is now too much happening in this PR. This would need to be split and squashed.

PR #354, opened 2023-05-18
galp6,galp7: Enable firmware security

crawfxrd:
Enable firmware security for galp7. Because the board is shared with galp6, that gets enabled as well.


crawfxrd:
If we don't want to enable it on galp6, the build script in firmware-open can be modified so it can be added to `models/galp7/ec.config`.

jackpot51 (approved):

leviport (approved):
Working well on galp6 and galp7

PR #355, opened 2023-05-19
Add reset command to ectool

13r0ck:


tool/src/main.rs
```diff
@@ -601,6 +602,13 @@ fn main() {
                 process::exit(1);
             },
         },
+        Some(("reset", _sub_m)) => match unsafe { ec.reset() } {
+            Ok(()) => (),
+            Err(err) => {
+                eprintln!("failed to save LED settings: {:X?}", err);
```
crawfxrd, line 608: Message needs update

jackpot51, line 608: Wrong error message here.

13r0ck: thanks

13r0ck: thanks

jackpot51 (dismissed):

tool/src/main.rs
```diff
@@ -601,6 +602,13 @@ fn main() {
                 process::exit(1);
             },
         },
+        Some(("reset", _sub_m)) => match unsafe { ec.reset() } {
+            Ok(()) => (),
+            Err(err) => {
+                eprintln!("failed to send reset command: {:X?}", err);
```
crawfxrd, line 608: Is it only *sending* the command that will fail? What about: "failed to reset device"

crawfxrd (approved):

XV-02 (approved):
Brock mentioned that this was for exposing the Launch bootloader. Running it with the command he indicated (`ec --access hid reset`) worked, and running it haphazardly elsewhere didn't break anything, so looks good to me.

PR #358, opened 2023-05-25
galp{6,7}: Reduce PL4 on AC to 65W

crawfxrd:
Reduce PL4 to the same value as other non-GPU units, even though the galp7 has an H series CPU.

### DIFF
--- a/src/board/system76/galp6/board.mk
+++ b/src/board/system76/galp6/board.mk
@@ -37,7 +37,7 @@ CFLAGS+=\
 
 # Set CPU power limits in watts
 CFLAGS+=\
-	-DPOWER_LIMIT_AC=90 \
+	-DPOWER_LIMIT_AC=65 \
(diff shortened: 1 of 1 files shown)


leviport:
This fixes the USB-C monitor overdraw bug I was seeing. With my USB-C power meter, I'm seeing an average draw of ~76W over USB-C while building the kernel.

Does dropping the limit to 65W negatively affect performance though? It'd be a shame if we had to leave some of those H-class gains on the table.

crawfxrd:
I simply don't know enough about measuring and configuring power limits, plus how all the values interact to make more in depth changes though.

Better to throttle it before release to avoid issues like I did with other ADL/RPL boards, and then work on improving power limits later (such as enabling PMON and configuring Psys values).

leviport (approved):
Sounds reasonable to me

13r0ck (approved):

PR #360, opened 2023-05-31
bonw15: invert USB power GPIO

jackpot51:
Fixes back left USB-A port

### DIFF
--- a/src/board/system76/bonw15/gpio.c
+++ b/src/board/system76/bonw15/gpio.c
@@ -65,7 +65,8 @@ void gpio_init() {
     GPDRC = BIT(6);
     // BL_PWM_EN_EC
     GPDRD = BIT(3);
-    GPDRE = 0;
+    // USB_PWR_EN# (inverted)
+    GPDRE = BIT(3);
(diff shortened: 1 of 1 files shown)


src/board/system76/bonw15/gpio.c
```diff
@@ -65,7 +65,8 @@ void gpio_init() {
     GPDRC = BIT(6);
     // BL_PWM_EN_EC
     GPDRD = BIT(3);
-    GPDRE = 0;
+    // USB_PWR_EN# (inverted)
```
crawfxrd, line 68: Inverted as in it's active high? Should we change it to `USB_PWR_EN` and make a note on the GPIO config below?

jacobgkau (approved):
Both USB Type-A ports work with this change.

src/board/system76/bonw15/gpio.c
jackpot51: This is the name in the schematic, so I'm inclined to leave it.

crawfxrd (approved):

PR #361, opened 2023-06-01
scripts: Disable showing signature when getting date

crawfxrd:
Fixes building when `log.showSignature` is enabled.

See also: https://github.com/system76/firmware-open/pull/423

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ all:
 	@exit 1
 else
 # Calculate version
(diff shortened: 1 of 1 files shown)


jacobgkau (approved):
Tested this by enabling `log.showSignature` in my git config, creating a signed commit, then attempting to `make BOARD=system76/darp5`. Without the change, I got `Makefile:29: *** target pattern contains no '%'.  Stop.` With the change, the build executes as expected.

jackpot51 (approved):

PR #362, opened 2023-06-03
kbscan: Switch from GPIO mode to KBS (Normal) mode

crawfxrd:
Use the default mode for reading the keyboard scan matrix when being used as a keyboard. There should be no perceived change in behavior, but should make the code easier to understand.

Note: `KSO[17:16]` are configured by `GPRCRC` on boards that use them. They are now set to alternate function to use in KBS mode rather than GPIO mode, with the pull-up enabled to prevent them from floating when configured as open-drain.

As part of this change, we now only read the hardware matrix state once upfront, instead of on every iteration through the loop applying the logic.

Tested by cherry-picking to darp9 and verifying that typing still worked.

Need to test:

- [x] Wake from sleep on keyboard input
- [x] bonw14 (n-key rollover), but I don't think that has successfully booted on newer firmware in a while
- [x] the 14" keyboards
- [x] External debugging still works
- [x] typing *really fast*

TODO:

- [x] Don't re-read hardware state after initial scan (`kbscan_row_has_ghost()`)
- [x] Check if `kbscan_get_real_keys()` is even needed
  - I don't notice any difference

### DIFF
--- a/src/board/system76/addw1/gpio.c
+++ b/src/board/system76/addw1/gpio.c
@@ -105,11 +105,11 @@ void gpio_init() {
     // SMD_VGA_THERM
     GPCRC2 = GPIO_OUT;
     // KB-SO16
-    GPCRC3 = GPIO_IN | GPIO_UP;
+    GPCRC3 = GPIO_ALT | GPIO_UP;
     // CNVI_DET#_EC
     GPCRC4 = GPIO_IN | GPIO_UP;
     // KB-SO17
-    GPCRC5 = GPIO_IN | GPIO_UP;
+    GPCRC5 = GPIO_ALT | GPIO_UP;
     // PM_PWROK
     GPCRC6 = GPIO_OUT;
     // LED_ACIN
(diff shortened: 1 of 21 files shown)


crawfxrd:
Increases XRAM use by 18 bytes.
Decreases ROM use by 186 bytes.

```diff
--- build/system76/oryp8/0/ec.mem	2023-07-17 13:57:23.967001078 -0600
+++ build/system76/oryp8/1/ec.mem	2023-07-17 13:57:34.875145986 -0600
@@ -25,5 +25,5 @@
    Name             Start    End      Size     Max
    ---------------- -------- -------- -------- --------
    PAGED EXT. RAM                         0      256
-   EXTERNAL RAM     0x0001   0x051e    1310     2048
-   ROM/EPROM/FLASH  0x0000   0x8f8e   36218    65536
+   EXTERNAL RAM     0x0001   0x0530    1328     2048
+   ROM/EPROM/FLASH  0x0000   0x8ed4   36032    65536
```

leviport (approved):
Looks good on bonw14's NKRO keyboard, and it can keep up with as fast as I can type. I think this is working great.

crawfxrd:
May be better to use push-pull (default) rather than open-drain?
Faster slopes and lower power due to not using the pull-ups?

jackpot51 (approved):

PR #366, opened 2023-06-09
Add new keyboards for darp9 and oryp11

crawfxrd:
Add and use new keyboards for darp9 and oryp11.

TODO: Decide how to name them. I used part of the name printed on the sticker.


jackpot51:
I don't think this will work right when the keyboard configurator does things like resetting the keyboard. It would be better in the long run to just make it another keyboard type :-(

crawfxrd:
Wouldn't that mean darp9 needs its own keyboard as well?

Does keyboard-configurator need to know the keymap used?
Or can't it read the default keymap from the EC?

jackpot51:
The EC doesn't provide the default keymap. It has to have it hard-coded.

leviport:
After flashing darp9 with this, it looks like mic mute is working as expected, but the Keyboard Configurator no longer recognizes the keyboard as configurable. I think we should probably update the Keyboard Configurator alongside this, if we can.

XV-02:
On Oryp11, `Fn+F2` no longer toggles the display backlight, nor does that option appear in the keyboard configurator. However, it also isn't muting the mic, and that option is also (unsurprisingly) not showing in the keyboard configurator. 

What info do you need to trouble shoot this?

XV-02:
Nevermind. Rebooting the system got the mic mute working. I added popdev:master in there as well.

XV-02:
@leviport did this look good on Darp9? It looks good on Oryp11.

leviport:
Yeah it's good, but let's not forget to update the Configurator for it.

PR #368, opened 2023-06-21
peci: Clear status on completion

crawfxrd:
The PECI-over-eSPI implementation, like the legacy implementation, must acknowledge when the transaction is done by clearing the bit in the register.

Fixes a hang during shutdown on oryp11 after unplugging a TBT display.

TODO:

- Wait for any pending upstream transactions


jackpot51 (approved):

XV-02 (approved):
This is resolving the hang I was seeing on both shutdown and reboot. Approving.

PR #371, opened 2023-06-30
oryp11: Set adapter Rsense to 10 milliohms

crawfxrd:
PRS1 is 5, but it is in series with PRS3, which is also 5. PRS1 by itself is only connected to ACP/ACN through R331.

Ref: oryp11 schematics, AC In Charger (Sheet 59)


crawfxrd:
Other RPL boards need to be updated as well.

crawfxrd:
Also need to figure out wtf is going on with bonw15. PRS1 and PRS2 are 0.005 Ohms, but are in *parallel*.

jackpot51 (approved):

PR #376, opened 2023-07-17
Reset PS/2, clear KBC lock

crawfxrd:


src/board/system76/common/power.c
```diff
@@ -272,6 +274,10 @@ void power_on(void) {
     if (power_state != POWER_STATE_S0) {
         DEBUG("failed to reach S0, powering off\n");
         power_off();
+    } else {
+        kbc_clear_lock();
```
jackpot51, line 278: This is a strange place to put this.

src/board/system76/common/kbc.c
```diff
@@ -22,6 +22,11 @@ void kbc_init(void) {
     *(KBC.status) = BIT(4);
 }
 
+void kbc_clear_lock(void) {
```
jackpot51, line 25: I'd probably prefer a kbc_reset function to this function name.

crawfxrd:
KBC fix done in #377.

PR #377, opened 2023-07-17
Reset KBC on CPU reset

crawfxrd:
Should fix KBC being locked/disabled on boot.

TODO: Determine what is actually disabling it


leviport (approved):
I believe this is working. I don't know a reliable way to reproduce the unresponsive keyboard for certain, but I was seeing it frequently on bonw15. Previously, unplugging and re-plugging the AC adapter while the machine was off would make the keyboard work again. 

With this update, I did not see the keyboard become unresponsive in the firmware menu at all.

jackpot51 (approved):

PR #382, opened 2023-07-21
power: Set `XLP_OUT` at power off

crawfxrd:
Disable VDD3 once at power off instead of setting it every cycle in the LED update logic.

### DIFF
--- a/src/board/system76/common/power.c
+++ b/src/board/system76/common/power.c
@@ -317,6 +317,11 @@ void power_off(void) {
     // Configure WLAN GPIOs after powering off
     wireless_power(false);
 
+#if HAVE_XLP_OUT
+    // Power off VDD3
+    gpio_set(&XLP_OUT, 0);
+#endif // HAVE_XLP_OUT
+
     update_power_state();
 }
 
@@ -609,11 +614,6 @@ void power_event(void) {
             gpio_set(&LED_ACIN, !gpio_get(&LED_ACIN));
             last_time = time;
         }
-
-#if HAVE_XLP_OUT
-        // Power off VDD3 if system should be off
-        gpio_set(&XLP_OUT, 0);
-#endif // HAVE_XLP_OUT
     }
 
 //TODO: do not require both LEDs


crawfxrd:
Presumably this is the cause of https://github.com/system76/ec/pull/320#issuecomment-1396345712.

XV-02:
Does this need to be tested in tandem with https://github.com/system76/ec/pull/320#issuecomment-1396345712 ? 

In isolation, I'm seeing the power light flash amber for ~3 seconds or less when I hold the power-button to shut the machine off. I can't trigger the power light to flash by short-stroking the power button. 

All of these are resolutions of the issues I was seeing in https://github.com/system76/ec/pull/320#issuecomment-1396345712 . I'd need to check other systems before approving however. 

Is there any specific expected behaviour I should be looking for while I do that? (Its a very limited change, but I figure best to ask)

crawfxrd:
> Does this need to be tested in tandem with https://github.com/system76/ec/pull/320#issuecomment-1396345712 ?

No, I split this out to determine if this is the cause the incorrect LED behavior.

> In isolation, I'm seeing the power light flash amber for ~3 seconds or less when I hold the power-button to shut the machine off.

- Do you hold it after power off, or release it when it powers off?
- Does that happen on release firmware?

Sounds like something is being left on. Probably bad.

> Is there any specific expected behaviour I should be looking for while I do that?

One test is:

- Power off unit
- Remove AC power
- Apply AC power
- Ensure power LED does not turn on or blink


I will look into this more next week.

XV-02:
>     * Do you hold it after power off, or release it when it powers off?
> 
>     * Does that happen on release firmware?

So, I hold the button until the system powered down, and immediately release the button. It blinks there-after.

It does not happen on the currently released firmware.

XV-02:
>
> One test is:
> 
>     * Power off unit
> 
>     * Remove AC power
> 
>     * Apply AC power
> 
>     * Ensure power LED does not turn on or blink
> 

Doing this, I notice that, upon returning the AC power, the power light goes from off to solid amber, though this is also true on our current firmware release. 

A quick table of status light behaviour is steady-state situations. The results are the same for both released firmware, and with this PR. 


| AC |Battery State     | Power State| ----> | Power Status Light | Battery Status Light |
| ------------- | -------------- | ------------ | ---------- | ----------------------- | ------- |
| In   | Full |  On     |        |  Solid Green               | Solid Green  |
| In   | Full |  Suspend     |        |  Flashing Green               | Solid Green  |
| In   | Full |    Off   |        |  Solid Amber               | Solid Green  |
| In   | Partial |   On    |        |  Solid Green          | Solid Amber  |
| In   | Partial |  Suspend     |        |  Flashing Green               | Solid Amber  |
| In   | Partial |   Off    |        |  Solid Amber             | Solid Amber  |
| Out   | N/A  | On      |        |  Solid Green               | Off  |
| Out   | N/A |  Suspend     |        |  Flashing Green               | Off |
| Out   | N/A  |   Off    |        |  Off               | Off  |

I would also note that the behaviour between released firmware and this PR also differs when powering a system off with `shutdown now` while on AC:
- Released: Power light transitions from solid green to solid amber. Battery light remains as it was (amber for charging, green for full)
- This PR: Power and Battery lights turn off and stay off until AC is unplugged and plugged back in.

crawfxrd:
`XLP_OUT` controls VDD3, which is used for standby power. darp8 schematics show it's asserted at `AC_IN`. So it's not correct to set it after transitioning to S5. It has to be set when in S5 *and* AC is unplugged.

PR #385, opened 2023-07-24
Remove the use of symlinks

### LINKED ISSUE #383: Remove use of symlinks
Replace symlinks with real files. This will allow making model-specific changes.

Source files can handled with relative paths; e.g., `board-y = ../galp-3/gpio.c`.
Include directories still need to be handled, as they are included by the top-level Makefile.

```
$ find . -type l | sort
./src/board/system76/darp6
./src/board/system76/galp4
./src/board/system76/galp7
./src/board/system76/gaze16-3060-b
./src/board/system76/gaze17-3060-b
./src/board/system76/oryp10
```

crawfxrd:
Replace symlinks with directories with `board.mk`, using relative paths to include the files rather than duplicate them.

This allows making board-specific changes without affecting other boards, such as when firmware security had to be enabled on galp6 when it was enabled on galp7 (2d5cbadf7124).

Resolves #383.


XV-02 (approved):
These changes look fairly straightforward. The boards all built, and a couple flashed without issues. Calling it good.

jackpot51 (approved):

PR #386, opened 2023-07-24
power: Handle signals in separate functions

crawfxrd:


src/board/system76/common/power.c
```diff
@@ -599,6 +534,133 @@ void power_event(void) {
...
+        // Charging
+        gpio_set(&LED_BAT_CHG, true);
+        gpio_set(&LED_BAT_FULL, false);
+    }
+}
+#endif // HAVE_LED_BAT_CHG && HAVE_LED_BAT_FULL
+
+void power_event(void) {
+    const bool ac_new = power_handle_acin_n();
+    gpio_set(&AC_PRESENT, !ac_new);
```
crawfxrd, line 643: Does it need to be set on every cycle, or can it be set only when `ACIN#` changes?

src/board/system76/common/power.c
```diff
@@ -599,6 +534,133 @@ void power_event(void) {
...
+
+    // Configure charger based on charging thresholds when plugged in
+    if (!ac_new) {
+        battery_charger_configure();
+    }
+
+    // Read power switch state
+    const bool ps_new = power_handle_pwr_sw_n();
+    // Send power signal to PCH
+    gpio_set(&PWR_BTN_N, ps_new);
```
crawfxrd, line 653: Does it need to be set on every cycle, or can it be set only when `PWR_SW#` changes?

PR #387, opened 2023-07-24
kbscan: Return early if debouncing

crawfxrd:
A single timer is used for debouncing all keys, so there is no reason to perform any operations if the matrix is being debounced. Just return early and remove some of the convoluted logic.


XV-02 (approved):
As far as I can see, key debounce is working as desired still, and I'm not seeing any evidence of issues with using multiple layers either. Looks good.

jackpot51 (approved):

PR #393, opened 2023-07-31
peci: Check PECI available based on eSPI usage

crawfxrd:
If eSPI is used, use Virtual Wires to determine if PECI is available.

Fixes incorrectly reporting PECI as available on systems using S0ix, as the CPU would be in C10 but `PLTRST#` would not be asserted.

Requires enabling `HOST_C10` reporting in FSP-S: https://github.com/system76/coreboot/pull/189

Resolves: https://github.com/system76/firmware-open/issues/424

Test:

- [x] Suspend on a system using eSPI and S0ix (TGL-U)
- [x] Suspend on a system using eSPI and S3 (RPL)
- [ ] Suspend on a system using LPC and S3 (CNL)

### DIFF
--- a/src/board/system76/common/peci.c
+++ b/src/board/system76/common/peci.c
@@ -60,13 +60,6 @@ static struct Fan __code FAN = {
     .interpolate = SMOOTH_FANS != 0,
 };
 
-#if CONFIG_PECI_OVER_ESPI
-
-// Maximum OOB channel response time in ms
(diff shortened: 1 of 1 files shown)


crawfxrd:
Is this sufficient on systems using S3, or does `PLTRST#` still need to be checked? Do we need 3 versions of `peci_available()`?

jackpot51 (approved):

jackpot51 (approved):

PR #397, opened 2023-08-10
charger: Limit current based on battery capacity

crawfxrd:
Reduce the charge rate when the battery is near empty or near full to avoid potentially damaging the battery, such as causing swelling from overcharging.


crawfxrd:
This is effectively pointless with Clevo's requirement to stop charging at 90% of the SBS reported `FullChargeCapacity()`.

crawfxrd:
No clue how it impacts the battery.

Issue #391, opened 2023-07-29
Add support for IT5571E


crawfxrd:
It should be trivial to support the IT5571E used on the pang12, as it is nearly identical to the IT5570E used on most Clevo-based models.

This does not involve any changes to support the AMD CPU of the pang12.

PR #399, opened 2023-08-15
Add basic support for IT5571E

crawfxrd:
The IT5571E is nearly identical to the IT5570E.

Resolves: #391.


jackpot51 (approved):

leviport (approved):
I'm assuming there's no way to test this yet on pang12, so I'm approving based on a code review. We will do more extensive testing before anything rolls out.

PR #402, opened 2023-08-18
Add and use Make targets for deps/setup

crawfxrd:
Split the single `deps.sh` script into logically separate Make targets, including installing Rust in a separate script.

Note:

- CI works because GitHub runners have a thousand things installed, GNU Make included
- Submodules also aren't needed for CI as they are not used for building

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Install dependencies
-        run: ./scripts/deps.sh
+        run: make install-deps
 
       - name:  Check SPDX tags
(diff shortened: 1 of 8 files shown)


XV-02:
Are the CI checks from our github actions sufficient for testing CI functionality?

crawfxrd:
I'm just saying I'm not bothering to explicitly install `make` before using Make targets because GitHub runners are guaranteed to have it already.

XV-02 (approved):
As far as I can tell, the make targets are working as desired on Pop. Everything installed, and I was able to use EC tooling. I was able to build firmware with dependencies installed using the make target too, so that looks good.

PR #405, opened 2023-09-07
kbscan: Increase debounce time from 5ms to 10ms

crawfxrd:
Some users are reporting that a debounce time of 5ms is not enough to prevent keys from registering twice. Split the difference between the old and the new debounce times and set it to 10ms.

Ref: https://github.com/system76/firmware-open/issues/471

### DIFF
--- a/src/board/system76/common/kbscan.c
+++ b/src/board/system76/common/kbscan.c
@@ -21,7 +21,7 @@
 #endif // KM_NKEY
 
 // Debounce time in milliseconds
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

leviport (approved):
I have two darp8's set up side-by-side, one with 5ms and the other with 10ms, and I can't tell a difference, even typing ~115wpm on typing tests. I don't think my typing style is one that is affected by the 5ms debounce. It seems to be working though, so I approve.

PR #406, opened 2023-09-09
Fix sdcc error 283 - "function declarator with no prototype"

ilikenwf:
You may be using an older version, or perhaps recently set a -Werror type cflag, but the current master fails to build without this adjustment.

### DIFF
--- a/src/board/system76/common/pnp.c
+++ b/src/board/system76/common/pnp.c
@@ -41,7 +41,7 @@ void pnp_write(uint8_t reg, uint8_t data) {
     ec2i_write(0x2F, data);
 }
 
-void pnp_enable() {
+void pnp_enable(void) {
     DEBUG("Enable PNP devices\n");
 
     // Enable PMC1
--- a/src/common/include/common/version.h
+++ b/src/common/include/common/version.h
@@ -3,7 +3,7 @@
 #ifndef _COMMON_VERSION_H
 #define _COMMON_VERSION_H
 
-const char *board();
-const char *version();
+const char *board(void);
+const char *version(void);
 
 #endif // _COMMON_VERSION_H
--- a/src/common/version.c
+++ b/src/common/version.c
@@ -17,10 +17,10 @@ static const char __code VERSION[] =
     xstr(__FIRMWARE_VERSION__);
 // clang-format on
 
-const char *board() {
+const char *board(void) {
     return &BOARD[11];
 }
 
-const char *version() {
+const char *version(void) {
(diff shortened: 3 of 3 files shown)


jackpot51 (approved):

crawfxrd:
Please specify in the commit message the version of SDCC used.

ilikenwf:
> Please specify in the commit message the version of SDCC used.

I just updated the commit message. Hopefully it took - v 4.3.0 #14184

ilikenwf:
Just for future reference, I wonder what versions of compilers and libs are used when you're building official releases? I know that sometimes compiler differences can cause bricks...which may potentially be why I'm waiting for an arduino to arrive Monday because I jumped the gun playing with the firmware and EC code...and need the Arduino to allow the machine to charge and power on again after flashing a test build. I'm sure the UEFI firmware is probably fine but I think I must have screwed up the EC somewhere, or the compiler did.

Because I built on Arch, I'm wondering if I have "too new" when it comes to a compiler or lib, by chance?

Current versions on my build machine (not counting the deps that get compiled...which could also potentially have issues due to a newer compiler):

```
clang 15.0.7-9
curl 8.2.1-1
lib32-curl 8.2.1-1
python-pycurl 7.45.2-3
gcc 13.2.1-3
gcc-libs 13.2.1-3
gcc5 5.5.0-3
lib32-gcc-libs 13.2.1-3
automake 1.16.5-2
cmake 3.27.1-1
libpagemaker 0.0.4-3
make 4.4.1-2
pkgconf 1.8.1-1
systemd-libs 254.1-1
```

jackpot51:
> Just for future reference, I wonder what versions of compilers and libs are used when you're building official releases?

We use Ubuntu Server 22.04, using `scripts/deps.sh` to install dependencies.

ilikenwf:
> > Just for future reference, I wonder what versions of compilers and libs are used when you're building official releases?
> 
> We use Ubuntu Server 22.04, using `scripts/deps.sh` to install dependencies.

Thank you!

I suppose it may be wise for me to setup a VM and build there just for safety's sake... I assume that a bad EC flash would indeed cause what I have with no button activity, lights, etc... this all happened after the flash, so I should just be able to build the previous stable tag and get it going again most likely...if not I may have to pay and overnight it to System76...

I've flashed plenty of bios chips, usually soic8, but I have, to be safe, a WSON adapter as well as the mega and adapter coming...I shouldn't have started playing until they arrived, I admit.

leviport (approved):

PR #415, opened 2023-10-12
Only support building a single board at a time

crawfxrd:
As a follow up to c461e20df364 ("make: Remove version from build output path"), only support building a single version of a single board by default. The build directory can still be specified by passing `BUILD`.

Fixes running the `clean` target when `BUILD` is a custom value or `BOARD` isn't set.

Fixes: b03c960b4f52 ("make: Reduce build output")

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,9 @@ ifneq ($(VERBOSE),1)
 MAKEFLAGS += -s
 endif
 
+# Set build directory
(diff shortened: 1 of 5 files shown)


jackpot51 (approved):

jacobgkau (approved):
Building on the CLI still works.

PR #416, opened 2023-10-12
time: Replace hard-coded values with macro

crawfxrd:
This is more descriptive than the comment and allows for changing values, such as using a chip with a different clock frequency.

The resulting binary is identical.

### DIFF
--- a/src/arch/8051/time.c
+++ b/src/arch/8051/time.c
@@ -3,23 +3,34 @@
 // Uses timer 0 to keep track of global time
 
 #include <8051.h>
-
 #include <arch/time.h>
+
+#define OSC_DIVISOR 12
+#define TICK_INTERVAL_MS 1
+// Value to reload into the timer when the overflow interrupt is triggered.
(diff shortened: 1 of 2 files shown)


src/arch/8051/time.c
```diff
@@ -3,38 +3,52 @@
 // Uses timer 0 to keep track of global time
 
 #include <8051.h>
-
 #include <arch/time.h>
 
+#define OSC_DIVISOR 12
+#define TICK_INTERVAL_MS 1
+// Value to reload into the timer when the overflow interrupt is triggered.
+#define TIMER_RELOAD (0xFFFF - ((TICK_INTERVAL_MS) * ((CONFIG_CLOCK_FREQ_KHZ) / OSC_DIVISOR)))
```
crawfxrd, line 11: Is this an off-by-one? Should it be subtracted from 0x1_0000 instead of 0xFFFF? This would cause `TL0` to be 0x02 instead of 0x01.

jackpot51: I'm not sure

jackpot51 (approved):

leviport (approved):
No regressions detected on a gaze16

PR #418, opened 2023-10-16
Add Containerfile for development

crawfxrd:
Add a command to create a Debian-based container with the SDCC and Rust toolchains installed in order to minimize the workspace setup time and ensure build environments are the same.

    make -C containers
    podman run -it --rm \
        -v $PWD:/workspace:Z \
        system76/ec:latest \
        make BOARD=system76/oryp8

A locally built image is 1.07 GiB.

### DIFF
--- a/containers/Makefile
+++ b/containers/Makefile
@@ -1,0 +1,22 @@
+# SPDX-License-Identifier: GPL-3.0-only
+
+# Disable built-in rules and variables
+MAKEFLAGS += --no-builtin-rules --no-builtin-variables
+.SUFFIXES:
+
+# Default to silent builds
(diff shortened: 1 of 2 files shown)


XV-02:
Should we update _ec/scripts/deps.sh_ to include the `podman` dependency for containerized builds?

crawfxrd:
There is no integration for using the container yet, so podman is not a dependency.

XV-02 (approved):
As far as I can see, EC is building correctly for every system I've tried using the containerized build process. Seems to be working if Engineering wants to move forward with it.

jackpot51 (approved):

leviport (approved):
Re-tested, still working

PR #421, opened 2023-11-30
Add CodeQL Workflow for Code Security Analysis

b4yuan:
# Summary

This pull request introduces a CodeQL workflow to enhance the security analysis of this repository.

# What is CodeQL

CodeQL is a static analysis tool that helps identify and mitigate security vulnerabilities. It is primarily intra-function but does provide some support for inter-function analysis. By integrating CodeQL into a GitHub Actions workflow, it can proactively identify and address potential issues before they become security threats.

For more information on CodeQL and how to interpret its results, refer to the GitHub documentation and the CodeQL documentation (https://codeql.github.com/ and https://codeql.github.com/docs/).

# What this PR does

We added a new CodeQL workflow file (.github/workflows/codeql.yml) that
- Runs on every pull request (functionality to run on every push to main branches is included as a comment for convenience).
- Runs daily.
- Excludes queries with a high false positive rate or low-severity findings.
- Does not display results for git submodules, focusing only on our own codebase.

## Validation

To validate the functionality of this workflow, we have run several test scans on the codebase and reviewed the results. The workflow successfully compiles the project, identifies issues, and provides actionable insights while reducing noise by excluding certain queries and third-party code.

## Using the workflow results

If this pull request is merged, the CodeQL workflow will be automatically run on every push to the main branch and on every pull request to the main branch. To view the results of these code scans, follow these steps:
1. Under the repository name, click on the Security tab.
2. In the left sidebar, click Code scanning alerts.

# Is this a good idea?

We are researchers at Purdue University in the USA. We are studying the potential benefits and costs of using CodeQL on open-source repositories of embedded software.

We wrote up a report of our findings so far. The TL;DR is “CodeQL outperforms the other freely-available static analysis tools, with fairly low false positive rates and lots of real defects”. You can read about the report here: https://arxiv.org/abs/2310.00205

# Review of engineering hazards

License: see the license at https://github.com/github/codeql-cli-binaries/blob/main/LICENSE.md: 

```
Here's what you may also do with the Software, but only with an Open Source Codebase and subject to the License Restrictions provisions below:

Perform analysis on the Open Source Codebase.

If the Open Source Codebase is hosted and maintained on GitHub.com, generate CodeQL databases for or during automated analysis, CI, or CD.
```

False positives: We find that around 20% of errors are false positives, but that these FPs are polarized and only a few rules contribute to most FPs. We find that the top rules contributing to FPs are: cpp/uninitialized-local, cpp/missing-check-scanf, cpp/suspicious-pointer-scaling, cpp/unbounded-write, cpp/constant-comparison, and cpp/inconsistent-null-check. Adding a filter to filter out certain rules that contribute to a high FP rate can be done simply in the workflow file.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,127 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main", "master" ]
+  #schedule:
+  #  - cron: '0 0 * * *'
+  pull_request:
+    branches: '*'
+
+jobs:
+  analyze:
+    name: Analyze
+    # Runner size impacts CodeQL analysis time. To learn more, please see:
+    #   - https://gh.io/recommended-hardware-resources-for-running-codeql
+    #   - https://gh.io/supported-runners-and-hardware-resources
+    #   - https://gh.io/using-larger-runners
+    # Consider using larger runners for possible analysis time improvements.
+    runs-on: ubuntu-22.04
+    timeout-minutes: 360
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'c-cpp' ]
+        # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+      with:
+        submodules: recursive
+
+    - name: Install dependencies
+      run: ./scripts/deps.sh
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v2
+      with:
+        languages: ${{ matrix.language }}
+        # If you wish to specify custom queries, you can do so here or in a config file.
+        # By default, queries listed here will override any specified in a config file.
+        # Prefix the list here with "+" to use these queries and those in the config file.
+
+        # For more details on CodeQL's query packs, refer to: https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs
+        # queries: security-extended,security-and-quality
+        queries: security-and-quality
+
+
+    # Autobuild attempts to build any compiled languages (C/C++, C#, Go, Java, or Swift).
+    # If this step fails, then you should remove it and run the build manually (see below)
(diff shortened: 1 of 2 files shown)


jacobgkau (commented):
Thank you for proposing some additional security CI.

The fourth commit is commenting out two of the lines that you added in the first commit. I see the same commit just commenting out those two lines in another repo where you opened a similar pull request: https://github.com/civetweb/civetweb/pull/1215 If you're going around submitting this to lots of repositories, it may be a good idea to just incorporate that change into the initial commit (or don't add those lines in the first place).

For that matter, why are the commit messages for the middle three commits identical? Repeating the multi-paragraph introduction four times seems a little overkill. If there's value in having the commits broken out, then each commit's description should be specific to what it's changing; if there's not, then it should just be a single commit. (I think we may have squashing via GitHub disabled in this repository, although I can't see right now since we don't have the reviews yet.)

Engineering can decide if this additional CI would be beneficial or not.

b4yuan:
@jacobgkau Thanks for the comments. Initial runs of CodeQL were done with those lines uncommented so that the workflow would be triggered on our fork per push. We decided that it would be better for an integrated workflow to be triggered daily and on every PR.

Our data collection step in the project was conducted months ago on >250 open source embedded repositories. We touch on this far more in depth in the paper: https://arxiv.org/abs/2310.00205, but we are now reaching out to devs with a fully configured CodeQL so that it may be of benefit to the repository.

.github/workflows/codeql.yml
```diff
@@ -74,8 +74,8 @@ jobs:
     #   If the Autobuild fails above, remove it and uncomment the following three lines.
     #   modify them (or add more) to build your code if your project, please refer to the EXAMPLE below for guidance.
 
-    - run: |
-        ./.github/workflows/codeql-buildscript.sh
+    - name: Build firmware
+      run: make list-boards | xargs -I'{}' make BOARD='{}' -j$(nproc)
```
crawfxrd, line 78: This won't work for building. And probably doesn't work for analysis either.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,126 @@
...
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'c-cpp' ]
+        # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
```
crawfxrd, line 41: This link has a list of supported compilers which (to no surprise) doesn't include SDCC. Does running CodeQL on this codebase even work?

szsam: It works, but only scans 11% of C source files.
image

szsam: Found some minor issues.
image

.github/workflows/codeql.yml
szsam: CodeQL monitors the build process and creates a database to represent the hierarchical structure of code in the repository.

.github/workflows/codeql.yml
crawfxrd: Right, so it only works with avr-gcc that we use for the Mega2560 dev tool.

crawfxrd:
To no surprise, SDCC with its custom formats and esoteric targets is not a supported compiler. CodeQL is not useful for this project.

PR #422, opened 2023-12-08
Add usbc_mux_info command and implement for oryp10

jackpot51:
This is to aid in https://github.com/system76/coreboot/pull/201


src/board/system76/common/usbpd/tps65987.c
```diff
@@ -253,3 +275,106 @@ void usbpd_enable_charging(void) {
...
+        WARN("usbc_mux_info does not support port %d\n", port);
+        return false;
+    }
+
+    uint8_t value[6] = { 0 };
+    int16_t res = i2c_get(&I2C_USBPD, USBPD_ADDRESS, REG_DATA_STATUS, value, sizeof(value));
+    if (res < 0) {
+        DEBUG("ERR %04X\n", -res);
+        return false;
+    } else {
```
crawfxrd, line 325: `else` block not needed; can reduce indentation by a level.

src/common/include/common/command.h
```diff
@@ -87,4 +89,37 @@ enum SecurityState {
...
+    // DP pin mode B
+    CMD_USBC_MUX_INFO_DP_MODE_B = (2 << 8),
+    // DP pin mode C
+    CMD_USBC_MUX_INFO_DP_MODE_C = (3 << 8),
+    // DP pin mode D
+    CMD_USBC_MUX_INFO_DP_MODE_D = (4 << 8),
+    // DP pin mode E
+    CMD_USBC_MUX_INFO_DP_MODE_E = (5 << 8),
+    // DP pin mode F
+    CMD_USBC_MUX_INFO_DP_MODE_F = (6 << 8),
```
crawfxrd, line 122: coreboot and chromeec use a bit per mode; should we do the same?

```c
/* DP Mode pin definitions */
#define MODE_DP_PIN_A BIT(0)
#define MODE_DP_PIN_B BIT(1)
#define MODE_DP_PIN_C BIT(2)
#define MODE_DP_PIN_D BIT(3)
#define MODE_DP_PIN_E BIT(4)
#define MODE_DP_PIN_F BIT(5)
```

jackpot51: That sounds good to me

PR #423, opened 2023-12-26
Allow controlling debug via `config.mk`

crawfxrd:
Add new Makefile configs for controlling the log level and mechanisms instead of having to modify the `CFLAGS` directly.

Allows adding values to top level `config.mk` instead of the nested `common.mk` file.

### DIFF
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -15,11 +15,19 @@ Terms used:
     ```bash
     make BOARD=system76/<model> console_internal
     ```
-1. If you're not seeing seeing expected output, check the
-  [`LEVEL` cflag][level_cflag]. This is an EC compile time configuration and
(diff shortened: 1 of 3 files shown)


leviport:
I'm not seeing a `config.mk` in the top level. Would a user be expected to generate their own if they want to use this new option?

crawfxrd:
Yes, the same way that `BOARD` or `KEYMAP` can be set. It can also be set in `ec.config` in firmware-open. This way I can just have a commented out config value and uncomment it when I want to enable logging, such as is done with coreboot and edk2.

leviport (approved):
I was tripped up by https://github.com/system76/ec/issues/424 and had to make USB images from firmware-open, but after that, everything seems to be working nicely with this PR.

PR #425, opened 2024-01-02
kbscan: Disable reading columns at init

crawfxrd:
The functionality is split into a separate function to optimize the use case for each.


jackpot51 (approved):

jacobgkau (approved):
The keyboard seems to still work normally (tested on darp7).

PR #426, opened 2024-01-02
kbscan: Do not read matrix if lid is closed

crawfxrd:
Pull the lid check out to the loop to avoid accessing the matrix when we know we do not need the data. It is left in kbscan (instead of simply disabling reading) to clear the state of the matrix data.

The lid check for wake is removed as it will never be true.

### DIFF
--- a/src/board/system76/common/kbscan.c
+++ b/src/board/system76/common/kbscan.c
@@ -65,11 +65,6 @@ void kbscan_init(void) {
 
 // Read the state of the row for the selected column.
 static inline uint8_t kbscan_get_row(void) {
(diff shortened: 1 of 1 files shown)


leviport (dismissed):
Tested on serw13. Keyboard is disabled when lid is closed, and enabled when lid is open.

jackpot51 (commented):
This may not properly release keys if pressed while the lid is being closed

jackpot51 (approved):

leviport (approved):
Keyboard still enables and disables as expected, and holding a key with a ruler while closing the lid makes that key get released when the lid is closed.

PR #428, opened 2024-01-03
main: Replace MOD with bitwise AND

crawfxrd:
This prevents `moduint` from being called for the superloop.

A bitwise AND is used instead of using the equivalent `% 4` as an indication that modulo should not be used, as not using a power of 2 (which is optimized) will result in an expensive call to SDCC library functions.


jackpot51 (approved):

leviport (approved):
serw13 seems the same after this change. Not seeing any issues.

PR #429, opened 2024-01-18
Conditionally compile USB-PD support

crawfxrd:
Boards may not have USB-PD. Remove the need for a "none" option by only adding USB-PD sources when enabled.

The resulting binary for boards with USB-PD enabled (addw3, bonw15, serw13) are identical. The binary for boards without USB-PD now have the empty calls optimized out.


XV-02 (approved):
I'm not seeing any regressions around this. On a Serw13, I see the same charge behaviour on a build which includes this as with our currently release firmware. On other systems, everything seems similarly hunky-dory. Approving.

jackpot51 (approved):

leviport (approved):

PR #430, opened 2024-01-18
Conditionally compile dGPU support

crawfxrd:


leviport (approved):
Tested on lemp10 and gaze18. Both seem good, and Nvidia is working fine on gaze18.

jackpot51 (approved):

PR #432, opened 2024-01-19
kbscan: Remove redundant column checks for ghosts

crawfxrd:
There is no need to check columns before the one specified as they will have already been checked in previous loops of the caller.


leviport:
It looks like this increases ghosting on the darp9 I tested it on. Mashing a bunch of keys causes ghost presses on keys that are far away from my hand and the keys I'm mashing. The currently released firmware doesn't behave the same way

Current firmware:
IMG_20240122_131128_393

This PR:
IMG_20240122_130718_926

PR #433, opened 2024-01-19
kbscan: Increase debounce time from 10ms to 20ms

uSpike:
I found that a debounce time of 10ms was causing very frequent twice-registered keys with my typing style.  After setting the debounce time to 20ms I find that no repeated characters occur when typing 10 line of text.

Fixes: https://github.com/system76/firmware-open/issues/509


leviport (approved):
This seems like it's working fine for me.

PR #434, opened 2024-01-24
Conditionally compile KBLED support

crawfxrd:
Remove the need for a "none" option by only adding KBLED sources when enabled.

The resulting binaries with KBLED enabled (all of them) are identical.


leviport (approved):
Looks good on gaze16-3050

PR #435, opened 2024-01-25
ite: Add support for 256K flash

crawfxrd:
The upcoming addw4 will use the IT5570E-256, which requires removing the assumption that every ITE chip is 128K. Introduce new configs so boards may select which flash size they use.


jackpot51 (approved):

XV-02 (approved):
This is a small, constrained change. It looks like it should be fine, so I'm going to approve.

PR #439, opened 2024-03-01
Mark pointers and pointed-to data as `const`

crawfxrd:
All binaries should be identical.


leviport (approved):
Binaries are identical

jackpot51 (approved):

PR #450, opened 2024-03-22
tool: Add error for write locked

crawfxrd:
Add a new error for the case of trying to flash when security is enabled and it is still locked and update the related docs.

### DIFF
--- a/docs/flashing.md
+++ b/docs/flashing.md
@@ -1,15 +1,27 @@
 # Flashing firmware
 
+## UEFI application
+
+The `flash.sh` script from the top-level firmware-open project will use
+firmware-update, the UEFI application which is used for normal system updates.
+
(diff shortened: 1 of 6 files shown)


jackpot51 (approved):

XV-02:
After running `ectool.sh security unlock` I get presented with the unlock prompt on every boot - hot or cold - until I cancel out of it. 

That doesn't feel correct, as I would expect it to prompt for a single boot, and then the standard locked behaviour to reassert on subsequent boots.

crawfxrd:
> After running `ectool.sh security unlock` I get presented with the unlock prompt on every boot - hot or cold - until I cancel out of it.

Addressed by https://github.com/system76/firmware-setup/pull/45.

XV-02 (approved):
This looks to be working as desired still, and all my concerns/thoughts have been addressed and reviewed.

PR #451, opened 2024-04-04
darp10: Add Darter Pro 10

crawfxrd:


jackpot51 (approved):

XV-02:
I believe this will impact EC as well:

It looks like we have the `Numpad *` key assigned to cycle backlight as a secondary function. As I understand it, the legend on the production keyboard will not have and backlight functions, so the default EC should not include that behaviour.

leviport (dismissed):

leviport (approved):

jackpot51 (approved):

PR #452, opened 2024-04-22
Remove GPB7 on IT5570E

crawfxrd:
On IT5570E, pin A1 is `VSTBY0` and not a GPIO pin.


jackpot51 (approved):

leviport (approved):
Tested on darp8. I saw no changes in behavior.

PR #456, opened 2024-05-01
acpi: Report RPM values instead of raw tachometer values

crawfxrd:
Calculate the fan RPM values in the EC instead of coreboot ACPI, as it depends entirely on the EC configuration.

Requires: https://github.com/system76/coreboot/pull/222


jackpot51 (dismissed):

jackpot51 (dismissed):

jackpot51 (approved):

crawfxrd:
Included in #474.

PR #457, opened 2024-05-03
peci: Perform offset calculation in temp function

crawfxrd:
Have `peci_get_temp()` return the actual temp instead of the offset, requiring the caller to make another calculation for the temp.


jackpot51 (approved):

crawfxrd:
Included in #474.

PR #461, opened 2024-05-16
Add lemp13-b

jackpot51:
Copy of lemp13 with lemp12 keyboard


crawfxrd (approved):

leviport (changes requested):
Fn + F2 should be mic mute. Layout is the same as darp10-b, not lemp12.
IMG_20240516_123204_503

leviport (approved):
All good now

PR #466, opened 2024-06-17
oryp12: Fix USB-PD config

crawfxrd:
Fixes: 1e4667f1d38e ("Conditionally compile USB-PD support")


leviport (approved):
I did a quick test on oryp12, and this builds, flashes, and behaves as expected.

Issue #442, opened 2024-03-01
Replace clang-format with uncrustify


crawfxrd:
We can't use clang/LLVM for anything due to it not supporting the 8-bit architectures we use (MCS-51, AVR), so we are installing 500+ MiB of dependencies for a C/C++ formatting tool.

Replace it with [uncrustify](https://github.com/uncrustify/uncrustify), which is only like 500 KiB.

There are _a lot_ of configs (800+), so getting to work about the same as clang-format make take some time to go through all the options.

- Alpine 3.19.2, Debian 13, Ubuntu 24.04: uncrustify 0.78.1 ([`config.txt`](https://github.com/uncrustify/uncrustify/blob/uncrustify-0.78.1/documentation/htdocs/config.txt))

Examples of projects using it:

- [neovim](https://github.com/neovim/neovim/blob/v0.9.5/src/uncrustify.cfg)
- [cppcheck](https://github.com/danmar/cppcheck/blob/2.13.4/.uncrustify.cfg)
- [edk2](https://github.com/tianocore/edk2/blob/edk2-stable202402/.pytool/Plugin/UncrustifyCheck/uncrustify.cfg)

PR #470, opened 2024-07-02
Replace clang-format with uncrustify

crawfxrd:
Resolves: #442


jackpot51 (approved):

leviport (approved):
I tested compilation on a half dozen various models, and saw no issues. I also flashed a lemp10 and did some quick tests, and I saw no issues with it.

Issue #472, opened 2024-07-03
IT8587E locks up when keyboard is enabled


Thulinma:
Hey there! I know this is technically an unsupported model, and an old one at that, but figured I'd at least report this here.
Ever since installing the open EC on my darp5, most of the system76_ectool commands never worked and/or would crash the EC (needing to disconnect the battery to restart it, as I found no other way to do so). I always figured this was just using "something" unsupported on my older laptop model, and didn't think much of it - I just didn't use the tool.
These days I'm running EC commit fc3bad29a2a31555bccaaacb6af6d20b7bb1b7f6 - and in the last few Linux kernels (at least on `6.9.7`) something seems to have changed that now causes the non-system76 ACPI kernel modules (I'm not sure which exactly - nor sure how I could find out) to do.... _something_ that also crashes the EC in the same way.
After some prodding, it looks like setting system76_acpi to load in early boot (by including it in the initrd image) prevents this crash from happening (likely because it prevents other ACPI drivers from probing and/or attempting to load..?).

Thankfully, I have a workaround (load the system76_acpi kernel module early) that keeps things stable - but I kinda feel like it should not be possible to crash the EC (or at least not this trivially easy)... So I'm reporting it here. ^_^

If there is anything I can do to help debug this, do let me know! I'm honestly a little confused that (apparently?) this doesn't affect other models and/or nobody else noticed/reported the problem. At least on my system, simply trying to run e.g. `system76_ectool security` with any argument will already cause an EC crash to happen.

Just to make clear what I mean with "EC crash", the symptoms are:
- Loss of keyboard
- Loss of power button
- Front LEDs no longer change in any way (stay on in their last state)
- Loss of fan (this is probably the most dangerous one, since...)
- The laptop **does** stay turned on! (Thankfully it does thermal throttling... 😌)
- Shutting down the laptop from the OS does _seem_ to work (besides any EC-controlled parts, like the LEDs, which stay in their previous state) but I can't easily tell if it's "really off" afterwards.
- Holding down the power button (even for several minutes uninterrupted) does _not_ reboot the EC or shut down the laptop at all.
- Trying to turn the laptop back "on" after a software-induced shutdown has no effect.
The only way out of this state I've found is disconnecting the battery and waiting a few seconds, then reconnecting it... which makes everything normal again.

[...]

PR #473, opened 2024-07-04
Enable power switch WDT on IT8587E boards

crawfxrd:
Enable PWRSW WDT 2 and use the default timeout of 10 seconds.

Allows forcing an EC reset in case it gets into an invalid state.

Models with IT8587E:

- addw1
- darp5
- darp6
- galp3-c
- galp4
- oryp5

Ref: #315
Ref: #472

### DIFF
--- a/src/board/system76/addw1/gpio.c
+++ b/src/board/system76/addw1/gpio.c
@@ -45,6 +45,9 @@ struct Gpio __code WLAN_PWR_EN =    GPIO(B, 0);
 // uncrustify:on
 
 void gpio_init(void) {
+    // PWRSW WDT 2 Enable
+    GCR8 = BIT(4);
+
     // Enable LPC reset on GPD2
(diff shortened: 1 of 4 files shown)


leviport (changes requested):
I flashed this on darp5, and it appears the EC continually resets itself every 10 seconds. If I power the machine on, it turns itself off. If I leave it plugged into AC, the two LEDs on the front stay orange, but they blink off and come back on every 10 seconds with no intervention. I'm trying to get some debugger output, but the debugger connection is being tricky again.

crawfxrd:
It is impossible to use the PWRSW WDT on IT8587E boards. Clevo connected `PWR_SW#` to `GPD0`/`RI#` and not `GPE4`/`PWRSW` on all of them; `GPE4`/`PWRSW` is connected to `DD_ON` instead.

PR #454, opened 2024-04-30
dgpu: Split out getting temp to a function

crawfxrd:
The idea is to separate the logic into two distinct tasks:

- Read temperature data
- Update fan duties

Ref: #390

Issue #390, opened 2023-07-29
Sync fans based on temp instead of duty


crawfxrd:
The current implementation syncs all fans to the highest duty.

Replace this with syncing to the highest temperature reported by the sensors.

This will allow independent curves for FAN0 and FAN1 that work as expected based on their actual points.

**Impact**: Low or Medium
**Rationale**: Most boards use the same table for both fans. In those cases, there will be no perceived change in behavior to the user. On boards that use a different table for each fan, the different in behavior and between fans may be perceivable to the user. (Looking at existing fan points, oryp9/10 will be the most significant.)

qnxdev:
I highly support this, my darp9 is super hot sometimes, I have to turn on fan manually with `Fn + 1` but why can't it do that  job?

When I use fan it cools down the laptop such that I can put on my lap & work.

leviport:
@qnxdev this is unrelated to what you're describing. The darp9 only has a single fan, and this issue is about fan speeds on models with dual fans.

swisscdn:
> @qnxdev this is unrelated to what you're describing. The darp9 only has a single fan, and this issue is about fan speeds on models with dual fans.

So shall I raise another issue for this?

leviport:
@swisscdn No need. User-customizable fan curves are already in the works.

PR #474, opened 2024-07-04
Support second fan without dGPU

crawfxrd:
- Fan logic in dgpu, peci modules is moved to fan module
- CPU (PECI) fan is renamed to FAN1
- dGPU fan is renamed to FAN2
- Boards must now define which PWM channel is used for each fan
- Boards must now always define their fan points
- darp10 can now have fan points for the second fan

Behavioral changes:

- ectool: Fans are now indexed from 1 instead of 0

Test on:

- Unit with 1 fan
- Unit with dGPU (2 fans)
- darp10 (2 fans, no dGPU)

Requires: system76/coreboot#222
Includes: #454, #456, #457, #481
Unblocks: #390


jackpot51 (dismissed):

jackpot51 (approved):

XV-02:
With this PR, how should ectool report fan speeds? 

Looking at a gaze16-3060-b as the dGPU system, it looked like ectool reported fan speed as an 8-bit value between 0 and 255 (inclusive). `sensors` reports rpm values which look correct, however.

crawfxrd:
`ectool fan` reads/writes from the duty cycle register (`DCRi`), so uses a value of 0-255.

XV-02 (approved):
The systems tested - lemp13, darp10, and gaze16-3060-b - all appear to be acting as expected. Approved.

PR #475, opened 2024-07-05
Initialize GPIOs from static config table

crawfxrd:
`gpio_init()` is now the same for every board, and only requires an array that defines the init data for GPIOs.

The idea is to be able to support tooling for generating and validating the GPIO configs.


jackpot51 (approved):

leviport (approved):
Looks good on lemp13-b

PR #477, opened 2024-07-05
Sync fans based on temp instead of duty

crawfxrd:
Fans may be different sizes, placed in asymmetrical positions, or have different amounts of venting through the chassis. These characteristics affect the ability for each fan, separately, to dissipate heat and generate noise.

Replace syncing fans to the highest duty calculated for each fan, based on separate thermal sensors, to using the highest reported temperature across all sensors to calculate each fan's duty for that highest temperature.

In other words: The old behavior synced fans based on the *output* value (duty), while this new behavior syncs fans based on the *input* value (temperature).

This allows tuning fans separately to better manage total system thermals and mitigate noise.

Tested by modifying FAN2 points on oryp6 to always run at 80% and then verifying that "CPU fan" spins up/down to different values than "GPU fan".

Resolves: #390.

### DIFF
--- a/src/board/system76/common/fan.c
+++ b/src/board/system76/common/fan.c
@@ -220,18 +220,15 @@ static uint8_t fan_get_duty(const struct Fan *const fan, int16_t temp) {
 }
 
 void fan_update_duty(void) {
-#if defined(FAN2_PWM) && !CONFIG_HAVE_DGPU
-    int16_t dgpu_temp = peci_temp;
+#if CONFIG_HAVE_DGPU
+    int16_t sys_temp = MAX(peci_temp, dgpu_temp);
+#else
+    int16_t sys_temp = peci_temp;
 #endif
 
-    uint8_t fan1_duty = fan_get_duty(&FAN1, peci_temp);
+    uint8_t fan1_duty = fan_get_duty(&FAN1, sys_temp);
 #ifdef FAN2_PWM
-    uint8_t fan2_duty = fan_get_duty(&FAN2, dgpu_temp);
-
-#if SYNC_FANS != 0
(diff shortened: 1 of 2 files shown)


XV-02:
Would we expect the fan with a steady-state fan-curve to boost above that duty cycle on initial spin-up?

To build firmware leveraging this, just to confirm, I should only need the `fan2` branch in firmware-open and this branch for ec, right?

crawfxrd:
> Would we expect the fan with a steady-state fan-curve to boost above that duty cycle on initial spin-up?

Maybe for a second while it ramps up? I'm not sure how long it would take to become stable.

> To build firmware leveraging this, just to confirm, I should only need the `fan2` branch in firmware-open and this branch for ec, right?

Correct.

XV-02 (approved):
The fan over-ramp seemed to last no more than 5 seconds before it settled down. I'd wondered if it was somehow a result of the Heat-up timing, as that's a 5 second sample period, or potentially a result of the fan-smoothing?

I was testing on a gaze16-3060-b with the fan duty values for Fan2 set at 80% for all temps over 60°C. After reaching steady state on Fan2, I saw Fan1 adjust to match the temps being reported by the system, including dipping below Fan2's duty cycle. 

That seems to be the desired behaviour. I think this is working well enough for approval.

I would like to see https://github.com/system76/firmware-open/pull/563 merged before this, but as far as I could tell, there were no critical issues with this branch and the current master for firmware-open.

crawfxrd:
Behavior on oryp9 or oryp10 should be tested as they have the most significant discrepancy between fan tables, and so may benefit from being adjusting as part of this change.

jackpot51 (approved):

PR #478, opened 2024-07-06
Update Rust toolchain config

crawfxrd:
Remove rust-src as it is not required. Use a minimal profile with clippy and rustfmt installed, instead of the default profile, as rust-docs is not needed.


jackpot51 (approved):

XV-02 (approved):
I'm not having any issues building EC with this PR. Approving.

PR #479, opened 2024-07-06
tool: Update clap to 4.5

crawfxrd:
I didn't check the docs. I just made it compile.


leviport (approved):
ectool builds and seems to be working

jackpot51 (approved):

Issue #480, opened 2024-07-08
Use a moving average for temperatures


crawfxrd:
Mitigate some fan noise by using a moving average instead of an instantaneous readings for thermals.

When starting a load, CPU will report rapid jumps in temp temperature. These spikes end up getting recorded by the cooldown table, and will cause the fans to run at higher than necessary duties for the actual loads.

PR #482, opened 2024-07-08
Use a moving average for temperatures

crawfxrd:
Mitigate some fan noise by using a moving average instead of instantaneous points for thermal logic.

When starting a load, CPU will report rapid jumps in temp temperature. These spikes end up getting recorded by the cooldown table, and will cause the fans to run at higher than necessary duties for the actual loads.

Current logic is set to update every 250ms, so use 4 points to average over 1s.

TODO:

- Evaluate interaction with `HEATUP`/`COOLDOWN`, and if that needs to change as well.

Resolves: #480

### DIFF
--- a/src/board/system76/common/dgpu.c
+++ b/src/board/system76/common/dgpu.c
@@ -9,6 +9,9 @@
 
 int16_t dgpu_temp = 0;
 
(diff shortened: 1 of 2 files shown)


crawfxrd:
Included in #500.

Issue #209, opened 2021-06-05
Interrupt-driven architecture


crawfxrd:
The current architecture of the firmware is a polling super loop. This may be the simplest architecture to implement and reason about, but it is not efficient or reliable.

An event-driven system allows us to:

- Handle events when they happen, not whenever we get to them
- Reduce CPU overhead: No longer checking conditions that may be false most of the time
- Put the EC in Idle mode for additional power savings

## Drawbacks

- Entire codebase must be evaluated for things such as:
    - Blocking IO
    - Non-reentrant functions
    - Critical sections and atomic access
    - Variable to be accessed in ISRs not marked `volatile`
- Possibility of an [interrupt storm](https://en.wikipedia.org/wiki/Interrupt_storm) degrading usability
- Difficultly of reproducing/debugging problems

## Details

The EC has the concept of "internal" and "external" events that trigger interrupts. All\* are routed through the 8051 `INT1#` line.

- Internal (INTC): Triggered by EC modules
- External (WUC): Triggered by GPIOs from the board

A potential path for implementing interrupt handling is:

- Move `INT1#` ISR from common code to EC code
- Acknowledge any interrupt the occurs to prevent a storm
    - IT5570E has a single non-maskable interrupt: INT159
- Enable EX1

INTC can now receive interrupts. All source are masked by default, so only the NMI will assert `INT1#` and call the ISR.

- Implement INTC logic
    - Should be common between both ITE ECs

Logic for enabling, disabling, configuring, and acknowledging INTC interrupts may be common between ECs. The IRQs are specific to each EC.

WUC sources are _per-board_: GPIOs correspond to different functionality. E.g., `LAN_WAKEUP#` is not available on lemp9/lemp10, `GPH4` on darp7, and `GPB2` on the rest of the boards.

- Implement mechanism for configuring external interrupts
    - Should be common between both ITE ECs
    - Must be able to associate a GPIO with:
        - WUC group and bit offset (for `WUESR*`/`WUEMR*` access)
        - If it should trigger on rising- or falling-edge
        - The corresponding INTC IRQ
        - The ISR to handle the event

### Time-triggered architecture

Strictly use timers as the interrupt source. Tasks would then be scheduled to run only at certain intervals.

Tasks that are considered critical or require a short turnaround time can be scheduled on the shortest interval.

PR #486, opened 2024-07-13
main: Run events at time intervals

crawfxrd:
Rewrite the main loop to run all its events at certain intervals of the systick instead of running most on every loop.

- `usbpd_event`: Every 4 cycles -> 1ms
- `power_event`: Every 4 cycles -> 1ms
- `kbscan_event`: Every 4 cycles- > 5ms
- `lid_event`: Every 4 cycles -> 500ms
- `board_event`: Every cycle -> 1ms
- `kbc_event`: Every cycle -> 1ms
- `pmc_event`: Every cycle -> 1ms
- `smfi_event`: Every cycle -> 1ms
- `battery_event`: Every 1000ms

Ref: #209, Time-triggered architecture

### DIFF
--- a/src/board/system76/addw1/board.c
+++ b/src/board/system76/addw1/board.c
@@ -5,8 +5,6 @@
 #include <board/kbc.h>
 #include <common/debug.h>
 
-extern uint8_t main_cycle;
-
 void board_init(void) {
     // Allow CPU to boot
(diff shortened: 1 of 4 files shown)


leviport (approved):
Tested on oryp12. Everything seems to still be working nicely.

jackpot51 (approved):

Issue #453, opened 2024-04-29
Replace hard-coded fan assumptions with configs


crawfxrd:
There are some hard-coded assumptions around fans that need to be addressed:

- There is 1 CPU fan
  - It uses PWM2 (`DCR2`)
- There is 1 dGPU fan
  - It uses PWM4 (`DCR4`)

darp10 is a unit without a dGPU, but has 2 CPU fans. They use PWM2 and PWM3.

It should be possible for a board to specify which interface each fan uses.

PR #489, opened 2024-07-18
darp10: Add custom fan points

crawfxrd:


XV-02:
If I'm reading this correctly, we're moving the fans between 27% and 77% duty cycles across the thermal range?

If that is correct, why are we capping at 77%?

crawfxrd:
It's harder to make our fan tables now by only looking at the Clevo tables since:

- They use a ladder (10 levels) instead of a linear scale
- They use target RPM instead of PWM duty
- They no longer document noise generated

Since we don't support target RPM, I used the average between 2 of the 6 profiles that Clevo provides as the duty.

- Fan1 max: 7700 RPM
- Performance, level 9: 7700 RPM
- Quiet, level 9: 5400 RPM

I don't know the actual duty required for the 7700 RPM target, so I considered it 100%.

jackpot51 (approved):

XV-02:
For the fan curves, where are we actually sampling the temps? are we using the "Package" temps or the "CPU" temp?

crawfxrd:
We use PECI to read the package temperature every 250+ ms.

- `src/board/system76/common/peci.c`
- `src/board/system76/common/main.c`

From the PECI user guide:

- The GetTemp() command is used to retrieve the maximum die temperature from a target PECI address.
- The processor temperature is averaged over 256 ms.

XV-02:
Okay, looks like the Max fan speeds are ~9000rpm +/- ~10% on the lab units. The duty cycle is working, and package/for temps are successfully driving fan speeds at the desired duty. 

Do you want to adjust the curves in light of different values for 100% duty than you expected? For reference, in a 25⁰C environment, building kernels and running stress loads aren't pulling temps much above 80⁰C, or fan speeds much above ~6200 rpm. If not, I'm comfortable to approve this.

crawfxrd:
No, I'll leave it as-is. The primary concern with moving away from the default points on all boards is having the fans cool below 70⁰C.

I want to look into #453 more which will change how all this is defined anyways.

XV-02 (approved):

PR #493, opened 2024-07-24
kbscan: Work around IT8587E hang

crawfxrd:
IT8587E is hanging when reading the keyboard matrix.

- Increasing the delay does not fix it
- Placing delays between `KSO*` writes does not fix it
- Code generated by SDCC looks valid to me

Rewriting it like this does fix it, although I don't know why.

Tested on darp5 with WDT enabled:

- WDT is no longer triggered
- Normal EC functionality like fans and powering off works again
- Keyboard input still works

Resolves: #472

### DIFF
--- a/src/board/system76/common/kbscan.c
+++ b/src/board/system76/common/kbscan.c
@@ -45,10 +45,21 @@ static inline bool matrix_position_is_fn(uint8_t row, uint8_t col) {
 // Assert the specified column for reading the row.
 static void kbscan_set_column(uint8_t col) {
     // Assert the specific bit corresponding to the column.
-    uint32_t colbit = ~BIT(col);
-    KSOL = colbit & 0xFF;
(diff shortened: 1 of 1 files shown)


leviport (approved):
This is working great on my darp5!

Thulinma:
Just installed/tested this, and I can confirm that on my laptop as well, this seems to resolve the crashes with no (noticeable) negative effects whatsoever. I tried running various commands that caused crashes on the previous build and booting into kernels that causes crashes, and none of them now result in a crash anymore. 👍
(Some commands do _fail_, e.g. `ec_tool security` and `ec_tool fan`, but none of them cause a crash anymore when before they did. Maybe I'm just using them wrong; the help built into the tool is fairly sparse on usage instructions.)

mmstick (approved):

PR #496, opened 2024-08-07
fan: Replace interpolation, smoothing with stepping

crawfxrd:
Fan noise is one of the top complaints reported. The existing
interpolation and smoothing logic has not sufficiently addressed the
issues with fans changing speeds too quickly in response to rapid
changes in thermals (particularly from PECI).

This behavior can be observed by with very basic tasks, such as:

- Powering on a system and logging into GNOME
- Starting a GUI application such as Firefox

Replace them with a fixed step update per event interval. Fans now have
a maximum amount they change change over time (3.9%/sec) as they move
towards a target duty.

This does **not** address fans constantly turning on/off at lowest duty point or providing cooling at higher duties to mitigate thermal throttling.

**Impact**: High
**Rationale**: If the rate of change is too slow it will not be able to handle a rapid rise in system temperatures, causing overheating. This may lead to sudden system power-off (`PROCHOT#`)  and damage to board components.

### Test

- Evaluate if duty rate of change is responsive enough to handle sustained loads like compilation or gaming
- Evaluate if noise is significantly reduced without compromising heat dissipation for short loads


garrettjwilke:
build fails for `gaze18`:
```
src/board/system76/gaze18/fan.c:7: syntax error: token -> ',' ; column 11
make: *** [src/arch/8051/toolchain.mk:52: build/board/system76/gaze18/fan.rel] Error 1
```

I was able to build by editing the `fan.c` and adding a `0`  in the blank space in the following line:
```
{ 60, , PWM_DUTY(0) },
```
never tested with this change though

crawfxrd:
You are attempting to build `wip/fan-rework`. The branch for this is `fan-step`.

garrettjwilke (approved):
built and tested on serw13, oryp12, lemp13, galp5/6

jackpot51 (approved):

PR #498, opened 2024-08-22
Save battery thresholds to flash

crawfxrd:
Make battery thresholds persistent by saving them in flash space. This allows configuring them after EC reset (system powered-off, unplugged) but before any system firmware or OS policy can be applied.

As part of this, the dynamic keymap is moved from the hard-coded address of 0x1FC00 to "the last sector"; Models with 256 KiB flash (`CONFIG_EC_FLASH_SIZE_256K=y`) will lose their layout and need to reconfigure it.

### Test

- When not configured, default values (90, 100) are used across multiple EC resets
- When configured, the profile/custom values are used across multiple EC resets
  - If flash is dumped, the values are present in the ROM image

Check values with:

- `system76-power charge-thresholds`
- `cat /sys/class/power_supply/BAT0/charge_control_{start,end}_threshold`

### DIFF
--- a/src/board/system76/common/acpi.c
+++ b/src/board/system76/common/acpi.c
@@ -221,6 +221,7 @@ void acpi_write(uint8_t addr, uint8_t data) {
 
     case 0xBD:
         battery_set_end_threshold(data);
+        (void)battery_save_thresholds();
         break;
 
 #if HAVE_LED_AIRPLANE_N
--- a/src/board/system76/common/battery.c
+++ b/src/board/system76/common/battery.c
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
 #include <board/battery.h>
+#include <board/flash.h>
 #include <board/gpio.h>
 #include <board/smbus.h>
 #include <common/debug.h>
@@ -12,6 +13,18 @@ uint16_t battery_charger_input_current = CHARGER_INPUT_CURRENT;
 // Default values to disable battery charging thresholds
 #define BATTERY_START_DEFAULT 0
 #define BATTERY_END_DEFAULT 100
+
+// Flash address to save charging thresholds to
+static const uint32_t BAT_CFG_ADDR = CONFIG_EC_FLASH_SIZE - (2 * 1024);
+static const uint16_t BAT_CFG_MAGIC = 0x4254;
+
+struct battery_config {
+    uint16_t magic;
+    uint8_t start_threshold;
+    uint8_t end_threshold;
+};
+
+static struct battery_config __xdata bat_cfg = { { 0 } };
 
 // Represents a battery percentage level, below which charging will begin.
 // Valid values are [0, 100]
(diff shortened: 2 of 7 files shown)


src/board/system76/common/battery.c
```diff
@@ -112,3 +125,38 @@ void battery_reset(void) {
...
+
+    if (bat_cfg.magic != BAT_CFG_MAGIC) {
+        bat_cfg.magic = BAT_CFG_MAGIC;
+        bat_cfg.start_threshold = BATTERY_START_THRESHOLD;
+        bat_cfg.end_threshold = BATTERY_END_THRESHOLD;
+        loaded = false;
+    }
+
+    battery_start_threshold = bat_cfg.start_threshold;
+    battery_end_threshold = bat_cfg.end_threshold;
```
jackpot51, line 143: These values should probably be tested for validity the same way they are in the ACPI code, by calling the battery_set_end_threshold function first and the battery_set_start_threshold function second.

src/board/system76/common/battery.c
```diff
@@ -13,6 +14,18 @@ uint16_t battery_charger_input_current = CHARGER_INPUT_CURRENT;
 #define BATTERY_START_DEFAULT 0
 #define BATTERY_END_DEFAULT 100
 
+// Flash address to save charging thresholds to
+static const uint32_t BAT_CFG_ADDR = CONFIG_EC_FLASH_SIZE - (2 * 1024);
```
jackpot51, line 18: The dynamic keymap has a hardcoded location of `0x1FC00`. I'd recommend to just use a hardcoded value of `0x1F800` here so they live next to each other regardless of the flash size, or adjust the dynamic keymap location to be CONFIG_EC_FLASH_SIZE - 1024;

src/board/system76/common/acpi.c
```diff
@@ -215,6 +215,7 @@ void acpi_write(uint8_t addr, uint8_t data) {
 
     case 0xBD:
         battery_set_end_threshold(data);
+        (void)battery_save_thresholds();
```
crawfxrd, line 218: I wasn't sure if I should make a command for saving the thresholds, to minimize the amount it could write to flash.

With this, I am expecting the values to always be changed as a pair and that the end threshold will always be changed second.

jackpot51: From the code it seems that we should make the end threshold be set first since setting the start threshold has a check on not exceeding the end threshold.

Maybe we need to set these atomically together.

jackpot51 (approved):

crawfxrd:
Writing to flash works.

```
0003f800: 54 42 56 5a ff ff ff ff ff ff ff ff ff ff ff ff  TBVZ............
```

Reading it back does not.

```
$ cat /sys/class/power_supply/BAT0/charge_control_{start,end}_threshold
1
1
```

And I don't know why.

jackpot51 (approved):

leviport (approved):
Tested on darp10-b, and this seems to be working great. The thresholds I set persist through multiple EC resets. Very nice!

Issue #492, opened 2024-07-24
Add support for AMD-based hosts


crawfxrd:
Add initial support for AMD-based hosts. Some of the work includes:

- Gate Intel-specific interfaces that are unconditionally included
- Add support for SB-TSI for reading host temp
- Refactor power module to handle different SoC flows

Also consider:

- Rework existing dGPU support for differentiating NVIDIA and AMD GPUs

Ref: [system76/ec@kudu6](https://github.com/system76/ec/tree/kudu6)
Ref: https://github.com/system76/firmware-open/issues/196

PR #499, opened 2024-08-28
Add platform config for selecting AMD/Intel host

crawfxrd:
Add a new `PLATFORM` config for selecting whether a board is Intel or AMD (or potentially even ARM64). Make PECI require Intel and fix building when selecting AMD.

SB-TSI is not implemented as part of this, which can be done as part of adding an AMD-based board (#254).

Ref: #492


jackpot51 (approved):

leviport (approved):
I compiled a bunch of different models, and they all went fine. Then I test flashed a lemp10 and it seems good.

PR #500, opened 2024-09-03
fan: Replace heatup/cooldown logic with up/down curves

crawfxrd:
Replace calculating the duty on every update with tracking the level in the fan tables to use. This allows knowing whether the system is currently heating up or cooling down.

Replace the heatup and cooldown tables with an expanded fan table that defines separate up/down curves. The separate cooldown curve allows running the fans a few degrees below the initial turn-on point so that they do not constantly turn on and off at the threshold.


crawfxrd:
#510 is a lot simpler to resolve the issue of fans constantly turning on/off at point 0 threshold.

PR #504, opened 2024-10-14
Fan Curve test update for galp5 and darp10-b

thomas-zimmerman:
Updates both the galp5 and darp10-b to a slightly less aggressive fan curve.


jacobgkau:
screenshot-2024-10-14-21-21-49

crawfxrd:
Changes to darp10-b need to be copied to darp10.

thomas-zimmerman:
I have this running my work galp5 and the fans react to thermal changes decently quickly. The curve seems good for the standard idle temp that this sits at (between 47 and 58c).

jackpot51 (approved):

jackpot51 (changes requested):
Will these units throttle before 100?

thomas-zimmerman:
On the galp5, I don't think I reach fan point above about 90c. stress-ng on all cores, I never get above 90c.

PR #505, opened 2024-12-04
Use distro provided rustup

crawfxrd:
This removes the need to download and run a shell script to set up the environment. The distro is instead trusted to provide a valid binary.

Fedora has a package called "rustup", but it really is just the rustup-init binary [\[1\]].

Additionally handle the behavior changes for rustup 1.28.0, which removes implicit toolchain installation but adds support for installing from `rust-toolchain.toml`.

[\[1\]]: https://src.fedoraproject.org/rpms/rustup

### DIFF
--- a/docs/dev-env.md
+++ b/docs/dev-env.md
@@ -20,11 +20,10 @@ toolset.
 
 A complete list of dependencies can be seen in `scripts/deps.sh`.
 
-### Cargo
+### Cargo (Rust)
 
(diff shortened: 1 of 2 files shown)


crawfxrd:
I suppose I should try to handle Debian Bookworm since it doesn't have rustup packaged.

crawfxrd:
Done in #519.

PR #507, opened 2024-12-14
tool: Restrict port IO interfaces to x86

crawfxrd:
Limit all interfaces that use port IO to x86 targets.

The tool probably needs a rewrite to handle AArch64. Too much assumes x86 interfaces.


crawfxrd:
Just do a rewrite if we ever get an AArch64 board with a DTS.

PR #508, opened 2024-12-26
Updated fans control: removed 48C point

vfdev-5:
I have Darter pro laptop with the following firmware:
```
board: system76/darp10
version: 2024-09-11_351d80d
```
When using the laptop there is an annoying fans behavior such that they start running and stops few seconds after running and repeat this pattern over and over. Checking the CPU temperature, seems like they start from 48C degrees and when CPU cools down to 47C degrees, then fans stop. This hypothesis was confirmed when I checked fan points in `src/board/system76/darp10/board.mk`.

In this PR, I removed the first fan point on 48C to stop that annoying fans behavior.

I reflashed the firmware on the laptop with the new ROM using this PR and there is no more such constant on/off fans behavior. Trying to see whether the next fan point wont lead to a similar problem...


crawfxrd:
This doesn't resolve the issue.

See #500 for a potential fix.

crawfxrd:
#510 should be sufficient for now.

vfdev-5:
I finally flashed the following schedule: https://github.com/system76/ec/compare/master...vfdev-5:ec:update-dart10-fans-control2 and this seems less disturbing.

PR #514, opened 2025-01-30
darp10-b: Make Fn+left/right act as Home/End

phcoder:
This is similar to some other laptops like chromebooks and lenovos while strictly speaking redundant, it's convenient to have access to all navigation buttons in the same corner


leviport:
I'm not sure about changing the default mapping on a single model. That sounds more like something that should be left to the end-user to configure on their own machine with the [Keyboard Configurator](https://github.com/pop-os/keyboard-configurator/).

If we do change this, it should probably be changed on all models that use the same keyboard (galp5-7, lemp9-12, lemp13-b).

phcoder:
> I'm not sure about changing the default mapping on a single model. That sounds more like something that should be left to the end-user to configure on their own machine with the [Keyboard Configurator](https://github.com/pop-os/keyboard-configurator/).
> 
> If we do change this, it should probably be changed on all models that use the same keyboard (galp5-7, lemp9-12, lemp13-b).

Changed to all models using this layout

jackpot51:
I think this generally makes sense. On Launch we also do Page Up/Down for Fn+Up/Down.

PR #517, opened 2025-02-25
kbc: Support get scancode command sequence

phcoder:
This improves compatibility with PS/2 protocol.

### DIFF
--- a/src/board/system76/common/kbc.c
+++ b/src/board/system76/common/kbc.c
@@ -155,6 +155,7 @@ enum KbcState {
     KBC_STATE_IDENTIFY_0,
     KBC_STATE_IDENTIFY_1,
     KBC_STATE_SELF_TEST,
+    KBC_STATE_SCANCODE_GET,
 };
 
 // TODO: state per KBC (we only have one KBC so low priority)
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

leviport (approved):
Tested on my oryp9. Not seeing any adverse effects.

crawfxrd:
oryp9 has psmouse blacklisted in system76-driver. Was psmouse loaded?

crawfxrd:
It's set by modprobe options. You can do `lsmod | grep psmouse` to check if it's loaded. It's probably best to test on a model that has it currently blacklisted (by removing it from the modprobe options and regenerating initramfs), because we _know_ it has issues.

Otherwise, anything in [`system76driver/products.py`](https://github.com/pop-os/system76-driver/blob/0353b7746156520eeb5097ca9d6c50645b08e92f/system76driver/products.py) with `remove_blacklist_psmouse` or without any `psmouse` action.

leviport:
I also gave this a test on darp10-b since it has `remove_blacklist_psmouse`. I confirmed that psmouse was not blacklisted since it showed up in `lsmod | grep psmouse`, and I ran some typing tests. Keyboard behavior seems perfectly normal with this change.

PR #519, opened 2025-03-11
scripts: Update `deps.sh`

crawfxrd:
- Use distro-provided rustup package on Fedora, Arch
- Use sh compatible commands where possible
- Update command for installing pinned Rust toolchain
- Remove progress messages
- Remove commands that conflict with running the script in other contexts
  - git-hooks breaks calling from firmware-open directory
  - submodule checkout is only required for external flashing and for generating initial board configs from firmware-open

### DIFF
--- a/scripts/deps.sh
+++ b/scripts/deps.sh
@@ -1,24 +1,17 @@
 #!/usr/bin/env bash
 # SPDX-License-Identifier: GPL-3.0-only
 
+# shellcheck disable=SC1091
+
 set -eE
 
-function msg {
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

jackpot51 (approved):

jacobgkau (approved):
Still works on Pop!_OS 22.04 & 24.04. New Arch & Fedora package names are correct. Now works when run from `firmware-open` directory where it didn't before.

Issue #443, opened 2024-03-01
boards: Replace use of `CFLAGS` with configs


crawfxrd:
Add configs for boards to replace the direct manipulation of `CFLAGS`.

For configs I have added, I have been using the prefix `CONFIG_` to match Kconfig.

The idea is to eventually have a proper tool for managing build configs, such as using Kconfig itself.

crawfxrd:
coreboot uses Kconfig imported from Linux with patches: [`util/kconfig`](https://github.com/coreboot/coreboot/tree/24.08/util/kconfig)

Zephyr appears to have a Python based implementation for parsing Kconfig files: [`scripts/kconfig`](https://github.com/zephyrproject-rtos/zephyr/tree/zephyr-v3.5.0/scripts/kconfig)

There are Rust bindings for Kconfig via tree-sitter: [tree-sitter-kconfig](https://github.com/tree-sitter-grammars/tree-sitter-kconfig)

PR #531, opened 2025-08-19
boards: Replace `CFLAGS` defines with Make variables

crawfxrd:
Move modifying `CFLAGS` to `common.mk` so that boards only set configs by Make variables.

The only remaining use of `CFLAGS` in board Makefiles is for including headers from another directory to minimize duplicate files for model refreshes.

This also adds the `CONFIG_` prefix to `CHARGER` and `KBLED`.

Test: Build with `VERSION=0` is identical.

Resolves: #443


garrettjwilke (approved):
built every firmware. all built correctly.

flashed and tested on a few laptops. all flashed correctly

PR #532, opened 2025-08-20
kbled: Merge I2C implementations

crawfxrd:
The only difference between the I2C implementations is the color channel layout. Introduce two new configs for boards to select which they use.

- `CONFIG_KBLED_I2C_CH_BGR_PACKED`: Channels are in `BBB|GGG|RRR` format
- `CONFIG_KBLED_I2C_CH_BGR_MIXED`: Channels are in `BGR|BGR|BGR` format

Resulting `ec.rom` binary is identical.


jackpot51 (approved):

jacobgkau (approved):
The `sha256sum` of `ec.rom` for oryp5 changed after rebuilding with this commit, but it also wasn't identical after going back to master and building again. I assume something else in the build process means it's non-reproducible.

I tried to flash this on oryp5 this morning and ran into a lot of problems getting a build/install to work on a fresh Pop!_OS 22.04 installation. This will end up getting tested when we do firmware update certification on any affected models, so I don't see a reason to hold the merge up for what looks like a simple change.

PR #544, opened 2025-11-24
Add devcontainer integration

crawfxrd:
Basic [devcontainer](https://containers.dev/) integration and config for VSCode.

Installed VSCode to make sure it works, but I don't use it. Don't know if adding other extensions would be useful, like a Makefile and C/C++ extension (esp. since we use SDCC dialect of C for 8051).

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,15 @@
+{
+  "name": "System76 EC",
+  "build": {
+    "dockerfile": "../tools/containers/ec/Containerfile"
+  },
+  "customizations": {
+    "vscode": {
+      "settings": {},
+      "extensions": [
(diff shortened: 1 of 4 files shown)


leviport (dismissed):

crawfxrd:
- SDCC toolchain is now provided as a separate container that must be built first.
- Fixes compiling SDCC 4.4.0 by using gcc-13 instead of default (gcc-14).
- Updates the base Debian image.

leviport (dismissed):

crawfxrd:
Debian Trixie has SDCC 4.5.0, so drop building 4.4.0 from source and use the distro-provided package.

Without having to build SDCC, the local Containerfile is viable to use without having to publish and maintain images.

leviport (approved):

jackpot51 (approved):

PR #547, opened 2025-12-09
Minor refactor of early init

crawfxrd:
Move early init logic out of app code and into SoC code.

- Architecture (8051)
  - It initializes Timer 0 as the systick; other SoCs may use a different configuration
- GCTRL,ECPM
  - These are ITE system block; other SoCs may not have something called "GCTRL"/"ECPM"


jackpot51 (dismissed):

jackpot51 (approved):

leviport (approved):
Tested on darp10-b, didn't see any issues.

PR #553, opened 2026-01-07
Move `board/system76/common` to `app`

crawfxrd:
Begin better separating application, SoC, drivers, and board support logic by moving the "common" code to a new directory.


garrettjwilke (approved):
all firmware for all systems build correctly with this change

PR #554, opened 2026-01-08
ec/ite: Move e-flash signature to 0x80

### LINKED ISSUE #518: SDCC 4.5.0 conflicts with ITE code signature, puts `HOME` segment at wrong address
On the current master (commit d15bc38de8e2c5eb28e2231701f56199385b6d65), building the `ec.rom` for the `darp10-b` results in a wrong size, where it is 262080 bytes instead of 262144 bytes.

## Steps to reproduce

1. Clone the repository and initialize submodules
2. Build with `make BOARD=system76/darp10-b`
3. Check the size of the `ec.rom` with, e.g.,  `wc -c build/ec.rom`

Disclaimer: I currently do not have another platform where I can try it out so I am not 100% sure if there is something wrong on my machine causing this issue.

crawfxrd:
According to the datasheet, the e-flash signature can be at any 16-byte aligned address in 0x40-0xF0. Because of an address conflict with SDCC 4.5.0, move the signature to 0x80. This matches the address for Zephyr RISC-V SoCs, like 82302.

Ref: https://github.com/zephyrproject-rtos/zephyr/pull/36379
Resolves: #518

### DIFF
--- a/src/ec/ite/signature.c
+++ b/src/ec/ite/signature.c
@@ -1,17 +1,34 @@
 // SPDX-License-Identifier: GPL-3.0-only
 
-#include <ec/espi.h>
+// ITE SoCs have a hardware mechanism to check if e-flash is programmed by a
+// 16-byte signature check at power-on. It must be at a 16-byte aligned address
+// at 0x40-0xF0.
 
+#include <stdint.h>
+
+// NOTE: SDCC 4.5.0 causes a conflict with the default address of 0x40, so the
+// signature is moved to 0x80, which matches the address used by RISC-V SoCs
+// in Zephyr.
+#define EFLASH_SIG_ADDR 0x80
+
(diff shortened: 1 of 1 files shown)


jackpot51 (dismissed):

jackpot51 (dismissed):

crawfxrd:
Must be checked on an IT8587 board. 32k clock selection changed from external crystal to internal clock generator.

There's no clock source connected to GPJ6/7 on any board and they're all configured as GPIO and not ALT function, so I don't think it would even work. I guess there's no issue because nothing is actually using the 32 kHz clock? (Proprietary firmware also has it set to use external crystal.)

jackpot51 (approved):

leviport (approved):
Tested on darp5. Seems fine.

PR #568, opened 2026-03-13
pnp: Fix init and document values

crawfxrd:
`IHIOA` notes that:

> The I/O address is an offset from the LSB bits of the address of the host controlled module.

So it should not be set to the actual addresses configured in `BADRSEL`, but the _offset_ of them. So 0 for the address port (0x2E) and 1 for the data port (0x2F).


crawfxrd:
I don't know what the implications of the `IHIOA` change are.
Does it matter since we don't enable resources in coreboot?

jackpot51 (approved):

leviport (approved):

PR #575, opened 2026-09-15
Improve fan behavior with linear interpolation and 4-point curves

thomas-zimmerman:
- fan: interpolation with linear 4-point fan curves
- fan: Disable EC fan reading, use fan_pwm command for duty control
- power: Disable EC fan reading, use fan_pwm command
- Update most boards to new fan curve
- Add thermal-test.sh for testing fan behavior

This set of changes have been tested on lemp12 hardware

These changes were authored with help AI from OpenCode using the Big Pickle model.


crawfxrd:
Only fan point changes were rebased. Scripts and app changes were lost.

